PR #7, opened 2019-06-07
Initial commit for artifactory-xray

michaelarichard:


modules/lead/toolchain/artifactory-xray.tf
```diff
@@ -0,0 +1,38 @@
...
+   length  = 10
+   special = false
+}
+
+//  data "helm_repository" "jfrog" {
+//   name = "jfrog"
+//   url  = "https://charts.jfrog.io"
+// }
+
+resource "helm_release" "xray" {
```
cplee: need to add a setting:

`server.service.type=ClusterIP`

server.service.type | Xray server service type | LoadBalancer
-- | -- | --

PR #25, opened 2019-06-19
Disable Logstash Plugins

Pactionly:
Comments out Logstash variables for plugins which are currently deemed unnecessary.

### DIFF
--- a/stacks/environment-aws/variables.tf
+++ b/stacks/environment-aws/variables.tf
@@ -36,8 +36,8 @@ variable "dashboard_version" {
   default = "0.2.0-eb0f89d7cf8"
 }
 variable "artifactory_license" {}
-variable "bitbucket_token" {}
-variable "jira_token" {}
+# variable "bitbucket_token" {}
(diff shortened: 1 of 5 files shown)


stacks/environment-aws/variables.tf
```diff
@@ -36,8 +36,8 @@ variable "dashboard_version" {
   default = "0.2.0-eb0f89d7cf8"
 }
 variable "artifactory_license" {}
-variable "bitbucket_token" {}
```
cplee: delete these lines instead of commenting out

cplee:
Also need to review:
https://github.com/liatrio/lead-terraform/blob/master/stacks/environment-aws/lead.tf#L60-L61 and https://github.com/liatrio/lead-terraform/tree/master/modules/lead/dashboard

stacks/environment-aws/variables.tf
Pactionly: Done

PR #29, opened 2019-06-26
Eng 771

jknight-liatrio:
Installing istio in the istio-system namespace via the helm tiller running with escalated rbac privileges in the lead-system namespace


jknight-liatrio:
I'm not sure this is really the right way to go about this. As documented in the ticket, there are significant problems with this.

PR #32, opened 2019-06-28
Eng 801

jknight-liatrio:


modules/common/istio/main.tf
```diff
@@ -23,6 +41,56 @@ resource "helm_release" "istio" {
...
+  }
+
+  set {
+    name  = "global.k8sIngress.enableHttps"
+    value = "true"
+  }
+
+  set {
+    name  = "global.k8sIngress.gatewayName=ingressgateway"
+    value = "true"
```
JonRudy: I believe this gatewayName is being set wrong, the name is supposed to be "global.k8sIngress.gatewayName" and the value itself is supposed to be "ingressgateway" or our ingress gateway name. Also not sure if these values need to be set up atm since we didn't have the cert manager fully working. Referenced from here -  https://istio.io/docs/tasks/traffic-management/ingress/ingress-certmgr/

modules/common/istio/variables.tf
```diff
@@ -1,2 +1,6 @@
 variable "namespace" {}
-variable "crd_waiter" {}
\ No newline at end of file
+variable "crd_waiter" {}
+variable "kiali_username" {
+  default = "admin"
+}
+variable "kiali_password" {}
```
cplee: this should be auto-generated as a terraform random string, then stuffed in a secret

stacks/environment-aws/variables.tf
```diff
@@ -38,6 +38,7 @@ variable "dashboard_version" {
 variable "artifactory_license" {}
 variable "slack_bot_token" {}
 variable "slack_client_signing_secret" {}
+variable "kiali_password" {}
```
cplee: remove

stacks/environment-local/variables.tf
```diff
@@ -31,6 +31,7 @@ variable "enable_xray" {
 variable "artifactory_license" {}
 variable "slack_bot_token" {}
 variable "slack_client_signing_secret" {}
+variable "kiali_password" {}
```
cplee: remove

stacks/environment-local/istio.tf
```diff
@@ -23,7 +23,8 @@ module "istio_system" {
   source             = "../../modules/common/istio"
   namespace          = "istio-system"
   crd_waiter         = "${null_resource.istio_init_delay.id}"
+  kiali_password     = "${var.kiali_password}"
```
cplee: remove

stacks/environment-aws/istio.tf
```diff
@@ -19,10 +19,12 @@ resource "null_resource" "istio_init_delay" {
     } 
     depends_on = ["helm_release.istio_init"] 
 }
+
 module "istio_system" {
   source             = "../../modules/common/istio"
   namespace          = "istio-system"
   crd_waiter         = "${null_resource.istio_init_delay.id}"
+  kiali_password     = "${var.kiali_password}"
```
cplee: remove

modules/common/istio/main.tf
```diff
@@ -6,6 +6,28 @@ module "istio_namespace" {
   }
 }
 
+esource "random_string" "kiali_admin_password" {
```
cplee: typo?

PR #39, opened 2019-07-10
ENG-859: Upgrade lead-terraform to Terraform 0.12

ltscif:


ltscif:
@cplee would like to get your thoughts on https://github.com/liatrio/lead-terraform/pull/39/commits/c0f9035c6b86c1c2eea159f101122037aca3f98e.  Not sure how much time we should spend looking for a workaround.

cplee:
@ltscif definitely want to avoid the unnecessary update from bcrypt.  any other ideas on how to ignore?

ltscif:
@cplee my recommendation would be to switch to `ignore_changes = [data]` because it doesn't appear that updating the config map will cause artifactory to reload its config without a restart.  This would be a stopgap until the issue is hopefully resolved in Terraform.  If needed documentation could be updating indicating the resource should be manually tainted if changes to the config are needed.  While this isn't ideal, I think the frequency of config updates will be less than or equal to the frequency of terraform executions.

The only other option I would think of is perhaps spending more time investigating whether a hashing of the password instead of the encryption of the password is a viable option for the value within artifactory `security.import.xml`.

cplee:
@ltscif agreed, let's go with `ignore_changes = [data]`

ltscif:
@cplee running into this....
```
Error: Reference to undeclared module

  on ../../modules/common/istio/outputs.tf line 10, in output "cert_issuer_name":
  10:   value = module.istio_cert_issuer.issuer_name

The configuration contains no module.istio_system.module.istio_cert_issuer.


Error: Reference to undeclared module

  on ../../modules/common/istio/outputs.tf line 14, in output "cert_issuer_dns_provider_name":
  14:   value = module.istio_cert_issuer.dns_provider_name

The configuration contains no module.istio_system.module.istio_cert_issuer.
```
Thought maybe the issue was that the values in `lead-terraform/modules/common/cert-issuer/outputs.tf` referencing `var` was causing a problem, but eliminated it.  Tried switching to and referencing a `local` but that didn't work either.  Even created a fake `kubernetes_namespace.test_ns` resource and changed the output to `kubernetes_namespace.test_ns.metadata[0].name` (which seems to be working for `module.istio_system.module.istio_namespace.name`), but that's not working either.  

It's interesting because when I comment out those outputs I get a successful plan, even references to `module.istio_system.module.istio_cert_issuer.helm_release.cert_manager_issuers` which clearly contradicts `The configuration contains no module.istio_system.module.istio_cert_issuer`.  Though it's `module.istio_system.module.istio_cert_issuer.helm_release.cert_manager_issuers will be destroyed`, so maybe that has something to do with it that I'm not immediately seeing.

Gonna come back to it later with a fresh set of eyes, but if you have a moment and wanna take a crack at it, would welcome a second pair of :eyes:.

ltscif:
@cplee Interesting.  I'm still seeing the above errors in my plan, and now an additional error since you added a `count` to the nginx_ingress resource....`values = [data.template_file.nginx_ingress_values.rendered]` needs an instance key now so it should be `values = [data.template_file.nginx_ingress_values[0].rendered]`.

PR #42, opened 2019-07-11
Eng 894 certs

JGuzman4:


modules/common/certificates/helm/certificates/templates/certificates.yaml
```diff
@@ -0,0 +1,22 @@
...
+  name: {{ Values.namespace }}-certificate
+  namespace: istio-system
+spec:
+  acme:
+    config:
+    - domains:
+      - {{ Values.cluster_domain }}
+      - *.{{ Values.cluster_domain }}
+      dns01:
+        provider: route53-liatrio
```
jphayes: I'm introducing an output in the istio module which will have this provider name so it's not hard-coded

modules/common/certificates/helm/certificates/templates/certificates.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: certmanager.k8s.io/v1alpha1
+kind: Certificate
+metadata:
+  name: {{ Values.namespace }}-certificate
+  namespace: istio-system
```
jphayes: There's an output from the istio module which has this namespace name

modules/common/certificates/helm/certificates/templates/certificates.yaml
JGuzman4: Is there a branch youre working on @jphayes ?

modules/common/certificates/helm/certificates/templates/certificates.yaml
```diff
@@ -0,0 +1,22 @@
...
+      - *.{{ Values.cluster_domain }}
+      dns01:
+        provider: route53-liatrio
+    commonName: {{ Values.cluster_domain }}
+  dnsNames:
+  - {{ Values.cluster_domain }}
+  - *.{{ Values.cluster_domain }}
+  issuerRef:
+    kind: Issuer
+    name: letsencrypt-dns
```
jphayes: I'm introducing an output in the istio module which will have this issuer name so it's not hard-coded

PR #44, opened 2019-07-11
ENG-860: add support to self-register with keycloak

ltscif:
Initial realm `Toolchain` created/updated on startup which includes self-registering capability.
This includes TF .12 from ENG-859.


ltscif:
Btw, I'm unable to test email sending through SES at this time since my `Developer` IAM role doesn't have the necessary permissions to create the SES smtp user.

modules/common/aws-ses-smtp/main.tf
```diff
@@ -0,0 +1,41 @@
+resource "random_pet" "ses_smtp" {
+  keepers = {
+    name = "${var.name}"
+  }
+}
+
+resource "aws_iam_user" "ses_smtp" {
+  name = "${var.name}-${random_pet.ses_smtp.id}"
+}
```
cplee: need to add `permission_boundary`

ltscif:
@cplee might wanna hold on merging this since it's tightly related to ENG-735/ENG-737.

PR #48, opened 2019-07-15
ENG-737: Integrate Gitlab with Keycloak

ltscif:


ltscif:
Merged in ENG-860 and ENG-735.

PR #50, opened 2019-07-19
Added role and role binding permissions for default Service Account

gesparza3:
This PR adds permissions for the default service account to get `pod` and `job` resources.

### DIFF
--- a/modules/lead/product/production.tf
+++ b/modules/lead/product/production.tf
@@ -130,7 +130,7 @@ resource "kubernetes_role" "default-production-role" {
   }
 
   rule {
-    api_groups = ["", "extensions"]
+    api_groups = [""]
     resources  = ["pods", "jobs"]
     verbs      = ["get"]
   }
--- a/modules/lead/product/production.tf
+++ b/modules/lead/product/production.tf
@@ -110,3 +110,60 @@ resource "kubernetes_role_binding" "jenkins_production_rolebinding" {
   }
 }
 
+resource "kubernetes_role" "default-production-role" {
+  provider = kubernetes.production
+  metadata {
+    name      = "default-production-role"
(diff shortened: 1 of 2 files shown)


modules/lead/product/staging.tf
```diff
@@ -110,3 +110,60 @@ resource "kubernetes_role_binding" "jenkins_staging_rolebinding" {
...
+    }
+
+    annotations = {
+      description = "Permission required for default Service Account to get pods and jobs in staging namespace"
+      source-repo = "https://github.com/liatrio/lead-terraform"
+    }
+  }
+
+  rule {
+    api_groups = ["", "extensions"]
```
cplee: too permissive

modules/lead/product/production.tf
```diff
@@ -110,3 +110,60 @@ resource "kubernetes_role_binding" "jenkins_production_rolebinding" {
...
+      "app.kubernetes.io/managed-by" = "Terraform"
+    }
+
+    annotations = {
+      description = "Permission required for default Service Account to get pods and jobs in production namespace"
+      source-repo = "https://github.com/liatrio/lead-terraform"
+    }
+  }
+
+  rule {
```
cplee: too permissive

modules/lead/product/staging.tf
cplee:
```suggestion
    api_groups = [""]
```

modules/lead/product/production.tf
```diff
@@ -110,3 +110,60 @@ resource "kubernetes_role_binding" "jenkins_production_rolebinding" {
...
+    api_groups = ["", "extensions"]
```
cplee:
```suggestion
    api_groups = [""]
```

PR #58, opened 2019-07-23
ENG-951: flagger support for istio

ltscif:


cplee:
How does slack integration work with many channels?  I'm thinking might make sense to disable the slack integration for now and we can have our slack operator monitor the canary CR for events.

ltscif:
There were a couple options I was looking into.  

1) Utilizing webhooks which are specified in the Canary CRD to post Slack messages specific to that Deployment. https://docs.flagger.app/how-it-works#webhooks
2) Submitting a Flagger PR which would allow for per Canary `Notifier` calls.  

Regardless of those options, I left the default Slack capabilities usable in case there was a desire to have a central Slack channel receiving all Canary deploys (from an admin standpoint), in addition to per Canary Slack channels (multi-tenant).

Getting the slack operator to monitor CR events might be another good option as well.  Let me know which direction you want to go.

cplee:
I was planning on the last thing there:
> Getting the slack operator to monitor CR events might be another good option as well. Let me know which direction you want to go.

ltscif:
Ok, I'll pull out the global slack config.  Did you want me to take a stab at the slack operator changes?  If so, can you point me to where that is?

ltscif:
Also, are we ok to use a global Prometheus server?  Should be fine if we have any others that are federating to the global one.  If not, let me know and I can work on a multi-tenant solution for that as well.

ltscif:
Note, the installed version of the Flagger helm chart requires kubernetesVersion: >=1.11.0-0.  Latest Docker Desktop (stable) has kubernetes 1.10.11.  If using Docker Desktop, you'll need to installed the edge release.

PR #59, opened 2019-07-23
adding flagger

JGuzman4:


cplee:
@JGuzman4 - can we combine this with #58

PR #66, opened 2019-07-30
cluster autoscaler tags

ChrisSchreiber:


cplee:
just to confirm...we are going back to the old tags, but with a newer version of helm chart?  safe to assume this was tested?

ChrisSchreiber:
> just to confirm...we are going back to the old tags, but with a newer version of helm chart? safe to assume this was tested?

Yep, tested this in the ignitelab-test account. The tags are set as a catch all so that the cluster-autoscaler should work regardless if the tags use kubernetes.io or k8s.io so it should be forward compatible with any changes.

The EKS module has kubernetes.io hardcoded as the tag name for the ASG so I thought it made sense to use that for our tags as well so they are consistent.

PR #69, opened 2019-08-03
Eng 964

ChrisSchreiber:
- Resources request optimizations 
- Fixes dependency issue for namespaces and range-limit
- Change product tf backend to s3

### DIFF
--- a/stacks/environment-aws/eks.tf
+++ b/stacks/environment-aws/eks.tf
@@ -82,7 +82,7 @@ module "vpc" {
   tags = merge(
     local.tags,
     {
-      "k8s.io/cluster/${var.cluster}" = "shared"
+      "kubernetes.io/cluster/${var.cluster}" = "shared"
     },
   )
 }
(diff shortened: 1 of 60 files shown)


stacks/environment-aws/eks.tf
```diff
@@ -212,6 +212,16 @@ resource "aws_iam_role_policy_attachment" "worker_ssm_role_attachment" {
   policy_arn = "arn:aws:iam::aws:policy/service-role/AmazonEC2RoleforSSM"
 }
 
+// TODO Remove once restricted permissions validated
+resource "aws_iam_role_policy_attachment" "s3_access_role_attachment" {
+  role = module.eks.worker_iam_role_name
+  policy_arn = "arn:aws:iam::aws:policy/AmazonS3FullAccess"
```
cplee: too permissive...can we lock this down to a specific S3 bucket?

Pactionly:
Removed S3 changes including permissions as part of switch to postgres backend.

PR #72, opened 2019-08-06
ENG-626: Support for running non-system components on spot instances

ltscif:


cplee:
Adding `WIP` label until after ignite-lab on Thursday

stacks/environment-aws/eks.tf
```diff
@@ -47,7 +48,24 @@ EOF
       protect_from_scale_in = true
       enabled_metrics       = ["GroupMinSize", "GroupMaxSize", "GroupDesiredCapacity", "GroupInServiceInstances", "GroupPendingInstances", "GroupStandbyInstances", "GroupTerminatingInstances", "GroupTotalInstances"]
       pre_userdata          = local.ssm_init
-    },
+      kubelet_extra_args    = var.enable_spot_instances ? "--node-labels=kubernetes.io/lifecycle=normal --register-with-taints=${var.ondemand_toleration_key}=true:NoSchedule" : ""
+    },   
+  ]
+
+  worker_groups_launch_template_mixed = [
```
cplee: should there be 3, one for each AZ?

Also, can you add `      enabled_metrics       = ["GroupMinSize", "GroupMaxSize", "GroupDesiredCapacity", "GroupInServiceInstances", "GroupPendingInstances", "GroupStandbyInstances", "GroupTerminatingInstances", "GroupTotalInstances"]
`  like you see on the ondemand clusters?

stacks/environment-aws/eks.tf
```diff
@@ -72,6 +90,15 @@ EOF
 data "aws_availability_zones" "available" {
 }
 
+data "template_file" "ondemand_toleration" {
```
cplee: does the toleration FORCE the schedule to use the ondemand instance?  seems like may node affinity might be needed?  or, anti-affinity on spot?

stacks/environment-aws/eks.tf
ltscif: Added `enabled_metrics`.  Also added spots across 3 AZs, in case any workloads that get run on spots require a PV/PVC tied to a specific AZ.

stacks/environment-aws/eks.tf
ltscif: Nice catch.  The toleration will just allow the pod to be scheduled on the tainted ondemand instances, but it could still be potentially scheduled onto a spot instance.  Added an affinity to this template to force the scheduling to an ondemand instance.

PR #76, opened 2019-08-07
Allow tiller-manager to access resources for canary deployments

gesparza3:
This PR gives permission to interact with Flagger and HPA resources to the `tiller-manager` role. This will allow users to define and perform Canary deployments.


gesparza3:
Currently allows all verbs for canary and HPA resources...Do you know of an easy way to determine which verbs are necessary?

PR #83, opened 2019-08-13
ENG-1024: ensure tiller doesn't get moved by cluster-autoscaler

ltscif:
This builds upon ENG-626 and using tolerations/taints defined there to ensure tiller is installed to an ondemand node and not a spot node.  Furthermore, cluster-autoscaler is disabled for the ondemand node group.


ltscif:
Merged into ENG-626 #72

PR #85, opened 2019-08-16
Set idle time for skaffold pod

jknight-liatrio:
This allows for a pod to stick around to allow jenkins to have an
executor that can perform slack feedback quickly


cplee:
Delay merging until next week, after DevOpsDays Dallas

PR #99, opened 2019-08-30
bump istio helm chart to 1.2.5

cplee:


mrparkers:
I assume this is no longer needed since we're on 1.4.2 now.

PR #100, opened 2019-09-04
Added linuxbrew compatibility

johnknapprs:
Issue: https://github.com/liatrio/lead-terraform/issues/101

Homebrew has [merged](https://github.com/Linuxbrew/brew#linuxbrew-has-been-merged-into-homebrew) with the Linuxbrew project.  This PR allows for Linux users who use homebrew to also take advantage of the Brewfile setup. 

In this case, casks are macOS specific.  This is why I changed `aws-vault` from a cask to a formula.  

I also updated the formula double quotes to single.  Brewfiles are _ruby_ based and single quotes are best practice.

### DIFF
--- a/Brewfile
+++ b/Brewfile
@@ -1,12 +1,9 @@
-cask_args appdir: "/Applications"
-tap "homebrew/cask"
-brew "terraform"
-brew "terragrunt"
-brew "kubernetes-cli"
-brew "kubernetes-helm"
-brew "aws-iam-authenticator"
-brew "azure-cli"
-brew "dnsmasq"
-brew "go"
-
-cask "aws-vault"
+# brew 'terraform'
+brew 'terragrunt'
+brew 'kubernetes-cli'
+brew 'kubernetes-helm'
+brew 'aws-iam-authenticator'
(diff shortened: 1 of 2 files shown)


Brewfile
```diff
@@ -1,12 +1,9 @@
...
-brew "terraform"
-brew "terragrunt"
-brew "kubernetes-cli"
-brew "kubernetes-helm"
-brew "aws-iam-authenticator"
-brew "azure-cli"
-brew "dnsmasq"
-brew "go"
-
-cask "aws-vault"
```
Pactionly: Replacing the `aws-vault` cask with a formula doesn't appear to work on macOS.

johnknapprs: Looks like y'all need multiple OS's for CI testing :P 

I've updated it more to have some more logic, but I'm not thrilled with having active scripts in a file that's just supposed to list packages. 

I've used scripts in the brewfile in the past with no issues, but it feels like bad practice.  

An alternative would be to have separate OS specific Brewfiles and include a note in the readme to use a file path flag.

PR #114, opened 2019-09-19
ENG 1195

Pactionly:


stacks/environment-aws/iam.tf
```diff
@@ -87,6 +87,71 @@ resource "aws_iam_role_policy" "external_dns" {
...
+            "Action": [
+                "autoscaling:UpdateAutoScalingGroup",
+                "autoscaling:TerminateInstanceInAutoScalingGroup",
+                "autoscaling:SetDesiredCapacity"
+            ],
+            "Resource": "*",
+            "Condition": {
+                "StringEquals": {
+                    "autoscaling:ResourceTag/k8s.io/cluster-autoscaler/enabled": "true",
+                    "autoscaling:ResourceTag/kubernetes.io/cluster/lead": "owned"
```
cplee: replace `lead` with `${var.cluster}`

Pactionly: Fixed

PR #116, opened 2019-09-20
ENG-1197

Pactionly:


stacks/environment-aws/iam.tf
```diff
@@ -205,8 +205,33 @@ resource "aws_iam_role_policy" "cluster_autoscaler" {
...
+  "Statement": [
+    {
+      "Effect": "Allow",
+      "Principal": {
+        "Federated": "${aws_iam_openid_connect_provider.default.arn}"
+      },
+      "Action": "sts:AssumeRoleWithWebIdentity",
+      "Condition": {
+        "StringEquals": {
+          "${replace(aws_iam_openid_connect_provider.default.url, "https://", "")}:sub": "system:serviceaccount:${var.system_namespace}:operator_jenkins"
```
cplee: this should be `toolchain_namespace`, not `system_namespace`

PR #120, opened 2019-09-25
Added gitops stage to pipeline

sjordan8:


Jenkinsfile
```diff
@@ -1,27 +1,53 @@
...
-            }
-            post {
-              success {
-                notifyStageEnd()
-              }
-            failure {
-                notifyStageEnd([result: "fail"])
-              }
-            }
+    stage('Gitops') {
```
cplee: this should only be for master branch

Jenkinsfile
```diff
@@ -1,27 +1,53 @@
...
+      post {
+        failure {
+          notifyStageEnd([result: "fail"])
+        }
+      }
     }
+  }
+}
+def version() {
+  return sh(script: 'git describe --tags --dirty', returnStdout: true).trim();
```
cplee: i don't think this will work because this isn't an actual tag...won't be able to be referenced from lead-environments

sjordan8: I don't think I understand, Where is the product version defined then?

ChrisSchreiber: Take a look at the `git describe` help especially what the --dirty argument does. I think the problem is that that git command creates a version kind of like the most recent tag but what we need in this case is the exact tag.

sjordan8: Why is this a problem only for lead terraform and not the other repositories we added the gitops stage to?

ChrisSchreiber: The other versions used in lead environment are used to create a docker image or helm chart and then referenced using that version number. It doesn't actually matter what the version is as long as it is unique and using a semver makes it easier to track. 

The product version is used by terragrunt to checkout the source repo to run terraform apply against so it has to be an actual branch, commit or tag to clone the repo.

sjordan8: I think I fixed it, thanks for the great explanation 🤜 🤛

cplee:
@sjordan8 I still don't think version is correct.  Let's chat about how to work around it.

PR #128, opened 2019-10-09
Fix path to grafeas helm chart

jallen2112:
Deploy grafeas to cluster

### DIFF
--- a/modules/lead/toolchain/grafeas.tf
+++ b/modules/lead/toolchain/grafeas.tf
@@ -21,8 +21,7 @@ module "certificate" {
 resource "helm_release" "grafeas" {
   name       = "grafeas-server"
   namespace  = var.namespace
-  repository = "../../modules/lead/toolchain"
-  chart      = "grafeas-chart"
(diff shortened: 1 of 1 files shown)


modules/lead/toolchain/grafeas.tf
```diff
@@ -21,7 +21,7 @@ module "certificate" {
 resource "helm_release" "grafeas" {
   name       = "grafeas-server"
   namespace  = var.namespace
-  repository = "./grafeas-chart/"
+  repository = "../../modules/lead/toolchain"
```
cplee: why is this path going up and then back down to same path?

jallen2112: The working directory when terraform tries to deploy the grafeas helm chart is stacks/environment-aws, which threw us off when we were trying to get this working because we also assumed that the working directory would be modules/lead/toolchain

cplee: Check out this sample: https://github.com/liatrio/lead-terraform/blob/master/modules/lead/infrastructure/certmanager.tf#L5

PR #133, opened 2019-10-14
Eng 1261

jallen2112:
Deploy Grafeas server to cluster


modules/lead/toolchain/variables.tf
```diff
@@ -33,6 +33,10 @@ variable "ingress_external_traffic_policy" {
   default = ""
 }
 
+variable "grafeas_version" {
```
cplee: Can you pull this all the way up into `environment-aws` so that we can specify the version in the lead-environments repo?  Will also need to update the pipeline for the helm chart to trigger a gitops PR when the grafeas version changes.

jallen2112: fixed

PR #136, opened 2019-10-16
Eng 1205

JonRudy:
Add in helm charts to deploy kritis into the cluster and use grafeas as a backend.


modules/lead/toolchain/grafeas.tf
```diff
@@ -18,7 +18,7 @@ module "certificate" {
   enabled = true
   name = "grafeas-cert"
   namespace = var.namespace
-  domain = var.root_zone_name
+  domain = "grafeas-server.toolchain.svc"
```
cplee: The namespace name `toolchain` is configurable via `var.namespace`...can you use that instead?

modules/lead/toolchain/kritis.tf
```diff
@@ -0,0 +1,65 @@
+module "kritis_certificate" {
+  source = "../../common/certificates"
+
+  enabled = true
+  name = "kritis-cert"
+  namespace = var.namespace
+  domain = "kritis-validation-hook.toolchain.svc"
```
cplee: The namespace name toolchain is configurable via var.namespace...can you use that instead?

JonRudy: Fixed to use namespace var.

modules/lead/toolchain/grafeas.tf
JonRudy: Fixed to use namespace var.

PR #142, opened 2019-10-25
Certificate rbac

ChrisSchreiber:


modules/common/certificates/helm/certificates/templates/rolebinding.yaml
```diff
@@ -11,9 +11,9 @@ metadata:
...
 roleRef:
   apiGroup: rbac.authorization.k8s.io
-  kind: Role
-  name: {{ .Release.Name }}
+  kind: ClusterRole
+  name: lead-system-certificate-watch
 subjects:
 - kind: ServiceAccount
-  name: {{ .Release.Name }}
+  name: default
```
cplee: i'd rather not use the `default` service account.  the pattern that we've used elsewhere is to create the sa, role and role binding in terraform, and then pass the sa to the helm chart as a value...can we go with that pattern for consistency and least privilege?

mrparkers:
@cplee I pushed up some changes to use a dedicated service account created in the namespace module.

I don't have this running locally yet so I'd appreciate it if someone could verify this works for them while I work on my local setup.

One thing I noticed while working on this was that the production and staging certificates don't use `wait_for_cert = true`, meaning that the job we're trying to fix shouldn't even be running in the first place.  So I'm not sure why we're receiving terragrunt errors when configuring the production and staging namespaces.

modules/common/namespace/main.tf
```diff
@@ -97,3 +97,30 @@ resource "kubernetes_role_binding" "tiller_role_binding" {
...
+resource "kubernetes_role_binding" "certificate_watcher_role_binding" {
+  count = var.enabled ? 1 : 0
+  metadata {
+    name      = "certificate-watcher-binding"
+    namespace = kubernetes_namespace.ns[0].metadata[0].name
+  }
+  role_ref {
+    api_group = "rbac.authorization.k8s.io"
+    kind      = "ClusterRole"
+    name      = "certificate-read"
```
mrparkers: I wanted to use variable interpolation here to reference the name of the `ClusterRole`:

https://github.com/liatrio/lead-terraform/blob/75afa96b17b88893fa4a8d58ad8db0d13bae5a36/modules/lead/infrastructure/certmanager.tf#L58

But I wasn't able to figure out a good way to do that.

cplee: why not move these out of the namespace module and back into the certmanager.tf

mrparkers: I believe the service account has to be created within the namespace, otherwise the job in that namespace wouldn't be able to use it.

Since the service account needs to stay here, I think it would be easiest to keep the `RoleBinding` here as well so we can easily interpolate the variables for the service account name.

PR #146, opened 2019-10-30
fix: disable keycloak and mailhog locally, fix product variable name

mrparkers:


local/environment/terragrunt.hcl
```diff
@@ -28,8 +28,8 @@ inputs = {
   enable_gitlab      = false
   enable_istio       = false
   enable_grafeas     = false
-  enable_keycloak    = true
-  enable_mailhog     = true
```
mrparkers: easier to run locally without keycloak and mailhog

stacks/product-local/variables.tf
```diff
@@ -36,7 +36,7 @@ variable "enable_istio" {
   default = true
 }
 
-variable "keycloak_enabled" {
```
mrparkers: this change is actually needed for sandbox and production to work

PR #165, opened 2019-12-05
Upgrade flagger

tnishida1:
* Upgrades flagger to 0.20.4
* Adds toggle to terraform for flagger CRD creation

### DIFF
--- a/modules/common/flagger/main.tf
+++ b/modules/common/flagger/main.tf
@@ -1,6 +1,6 @@
-# Create CRDs for the flagger
+# Create CRDs for flagger
 resource "helm_release" "flagger_crds" {
-  count = var.enable ? 1 : 0
+  count     = var.enable ? 1 : 0
   name      = "flagger-crds"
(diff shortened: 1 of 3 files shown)


mrparkers:
note: we have to taint the current flagger helm release in terraform and manually delete the old CRDs before applying this change to a cluster.

modules/common/flagger/main.tf
```diff
@@ -1,3 +1,22 @@
+# Create CRDs for the flagger
```
mrparkers:
```suggestion
# Create CRDs for flagger
```

modules/common/flagger/main.tf
```diff
@@ -1,3 +1,22 @@
...
+resource "helm_release" "flagger_crds" {
+  count = var.enable ? 1 : 0
```
mrparkers:
```suggestion
  count     = var.enable ? 1 : 0
```

PR #170, opened 2019-12-10
Pinning helm to v2 (currently 2.16.1)

RobertKelly:
Pinning helm to v2.  Otherwise, helm3 replaces it when upgraded.


RobertKelly:
https://github.com/Homebrew/homebrew-core/blob/master/Formula/helm@2.rb is the formula if interested

PR #176, opened 2019-12-12
Updating to allow public access to dashboard / newer grafana version

JonRudy:


modules/lead/toolchain/prometheus-values.tpl
```diff
@@ -1,3 +1,14 @@
+grafana:
+  grafana.ini:
+    auth.anonymous:
+      enabled: true
+      org_name: Main Org.
+      org_role: Viewer
+  image:
+    repository: grafana/grafana
+    tag: 6.5.1-ubuntu
```
ChrisSchreiber: Is there a reason we are using the ubuntu version?

JonRudy: Yes there was an error with the way alpine was handling dns lookup. Parker had found this issue here https://github.com/grafana/grafana/issues/20096

PR #191, opened 2019-12-26
adding persistent volume for prometheus

JonRudy:


modules/lead/toolchain/prometheus-values.tpl
```diff
@@ -43,6 +43,13 @@ prometheusOperator:
     enabled: false
 prometheus:
   prometheusSpec:
+    storageSpec:
+      volumeClaimTemplate:
+        spec:
+          accessModes: ["ReadWriteOnce"]
+          resources:
+            requests:
+              storage: 1Gi
```
sjordan8: is 1Gi enough storage for prometheus?

ChrisSchreiber: probably not... try it in sandbox and see how long it takes to run out of space?

PR #193, opened 2019-12-31
use variable for k8s storage class

mrparkers:


stacks/environment-aws/variables.tf
```diff
@@ -166,3 +166,7 @@ locals {
     "Cluster" = var.cluster
   }
 }
+
+variable "k8s_storage_class" {
+  default = "gp2"
```
ChrisSchreiber: This should probably be set in environment-local also

modules/lead/dashboard/dashboard-values.tpl
```diff
@@ -1,7 +1,7 @@
 elasticsearch:
 %{ if local == "false" }
   volumeClaimTemplate:
-    storageClassName: gp2
+    storageClassName: ${k8s_storage_class}
```
ChrisSchreiber: If you set this in environment-local you can move this out of the `if local` condition

mrparkers: the `else` condition also specifies a `volumeClaimTemplate` with some other configuration.  I think this will look messier if I move this out, since I'll need another `if` condition.  I can try it if you'd like.

PR #194, opened 2019-12-31
ENG-1439 update istio helm deployment to 1.3.6

jbankes:


modules/common/istio/istio-values.tpl
```diff
@@ -35,6 +37,7 @@ global:
       limits:
         cpu: 40m
         memory: 64Mi
+    protocolDetectionTimeout: 100ms
```
mrparkers: weren't there more updates in the values file?

jbankes: Yes but we don't have the parent stanzas in our tpl file at all so I opted to keep in line with what others before me did.

PR #197, opened 2020-01-02
fix harbor storageclass

mrparkers:


modules/lead/toolchain/charts/harbor-volumes/values.yaml
```diff
@@ -1,9 +1,12 @@
+storageClassName:
 components:
   registry:
+    size:
```
jbankes: Should this have a size declared?

mrparkers: The size is declared when we consume the helm chart in terraform. This is only to help my IDE with auto complete

jbankes: Awesome, thanks!

PR #202, opened 2020-01-06
Bumped up limits in istio system to the recommended amount

sjordan8:


modules/common/istio/istio-values.tpl
```diff
@@ -194,12 +203,3 @@ sidecarInjectorWebhook:
     limits:
       cpu: 25m
       memory: 64Mi
-
-security:
```
mrparkers: why remove the requests / limits for citadel?

sjordan8: I saw a duplicate of the security limits / requests so I removed one of them

mrparkers: ah, I didn't catch that.  did you want to modify the requests / limits for the one you kept?

PR #203, opened 2020-01-07
Eng 1411

jallen2112:
Integrate Kibana with Keycloak via a gatekeeper proxy. Create keycloak client for kibana. Enable tls for elasticsearch


modules/lead/dashboard/main.tf
```diff
@@ -24,4 +84,33 @@ resource "helm_release" "lead-dashboard" {
...
+  access_type = "CONFIDENTIAL"
+  standard_flow_enabled = true
+ 
+  valid_redirect_uris = [ 
+    "https://kibana.${var.namespace}.${var.cluster}.${var.root_zone_name}/oauth/callback"
+  ]
+ 
+}
+
+resource "keycloak_openid_audience_protocol_mapper" "audience_mapper" {
```
mrparkers: this resource should have a `count` that checks if keycloak is enabled

modules/common/certificates/certificate-values.tpl
```diff
@@ -2,3 +2,4 @@ domain: ${domain}
 issuer_name: ${issuer_name}
 altname: ${altname}
 waitForCert: ${wait_for_cert}
+namespace: ${namespace}
```
mrparkers: I don't think this is needed - the helm release usually defines the namespace.

modules/common/certificates/main.tf
```diff
@@ -6,6 +6,7 @@ data "template_file" "certificate_values" {
     issuer_name = var.issuer_name
     altname = var.altname
     wait_for_cert = var.wait_for_cert
+    namespace = var.namespace
```
mrparkers: same as above

PR #205, opened 2020-01-07
Updated resource requests/limits

sjordan8:


modules/common/istio/istio-values.tpl
```diff
@@ -172,11 +180,12 @@ mixer:
...
     resources:
       requests:
-        cpu: 64m
-        memory: 64Mi
+        cpu: 1000m
+        memory: 1G
       limits:
-        cpu: 128m
-        memory: 128Mi
+        cpu: 2000m
```
ChrisSchreiber: Are we sure about this? I know that this needs a big increase but 1 CPU request and 2 CPU limit seems like overkill

sjordan8: That's the recommended amount in the prometheus helm chart I believe

mrparkers: all of these are the "performance tuned" requests / limits that are in the Istio helm chart by default.

In general I don't really mind high limits for some of the core istio components, but since we aren't really using a lot of the features that mixer plays a part in, I'm fine with cutting these in half

modules/common/istio/istio-values.tpl
```diff
@@ -9,42 +9,50 @@ gateways:
...
       resources:
         requests:
-          cpu: 10m
+          cpu: 100m
           memory: 128Mi
         limits:
-          cpu: 100m
-          memory: 256Mi
+          cpu: 2000m
+          memory: 1024Mi
```
mrparkers: let's cut these limits in half too

modules/common/istio/istio-values.tpl
```diff
@@ -9,42 +9,50 @@ gateways:
...
     resources:
       requests:
-        cpu: 200m
+        cpu: 100m
         memory: 128Mi
       limits:
-        cpu: 300m
-        memory: 128Mi
+        cpu: 2000m
+        memory: 1024Mi
```
mrparkers: and these

modules/common/istio/istio-values.tpl
```diff
@@ -9,42 +9,50 @@ gateways:
...
-        cpu: 10m
-        memory: 32Mi
+        cpu: 100m
+        memory: 128Mi
       limits:
-        cpu: 40m
-        memory: 64Mi
-    protocolDetectionTimeout: 100ms
+        cpu: 2000m
+        memory: 1024Mi
```
mrparkers: we can cut these in half too

PR #213, opened 2020-01-14
Updated resource requests/limits for kiali and prometheus

sjordan8:


modules/common/istio/kiali-values.tpl
```diff
@@ -0,0 +1,9 @@
+kiali:
+  resources:
+    limits: 
+      cpu: 500m
+      memory: 128Mi
+    requests:
+      cpu: 30m
+      memory: 32Mi
+
```
mrparkers: requests and limits for kiali are set here: https://github.com/liatrio/lead-terraform/blob/master/modules/common/istio/charts/kiali/templates/deployment.yaml#L42-L48

PR #214, opened 2020-01-14
Added memory limits for tiller

sjordan8:


stacks/environment-aws/main.tf
```diff
@@ -35,6 +35,8 @@ provider "helm" {
     "spec.template.spec.affinity.nodeAffinity.requiredDuringSchedulingIgnoredDuringExecution.nodeSelectorTerms[0].matchExpressions[0].values[0]=preemptible",
     "spec.template.spec.tolerations[0].key=${var.essential_taint_key}",
     "spec.template.spec.tolerations[0].operator=Exists",
+    "spec.template.spec.containers[0].resources.limits.memory=512Mi",
+    "spec.template.spec.containers[0].resources.limits.memory=512Mi",
```
mrparkers: `256Mi` was working for me in sandbox, any reason this needs to be `512`?

sjordan8: That's a mistake, my bad

mrparkers: I think 256 is a good value for this though, that's what I had to bump it to in sandbox to avoid it getting OOMKilled

sjordan8: It's currently 512Mi in sandbox. Would you prefer 256? I made the mistake of setting the limit twice and not the request

mrparkers: Sorry, I meant 256 for requests, 512 for limits

sjordan8: Gotcha, I'll push up those changes right now

stacks/environment-aws/main.tf
```diff
@@ -35,6 +35,8 @@ provider "helm" {
     "spec.template.spec.affinity.nodeAffinity.requiredDuringSchedulingIgnoredDuringExecution.nodeSelectorTerms[0].matchExpressions[0].values[0]=preemptible",
     "spec.template.spec.tolerations[0].key=${var.essential_taint_key}",
     "spec.template.spec.tolerations[0].operator=Exists",
+    "spec.template.spec.containers[0].resources.limits.memory=512Mi",
+    "spec.template.spec.containers[0].resources.requests.memory=256Mi",
   ]
```
ChrisSchreiber: why no cpu request and limits?

PR #224, opened 2020-01-22
add persistence to jaeger

mrparkers:


.editorconfig
```diff
@@ -0,0 +1,14 @@
+# editorconfig.org
+root = true
+
+[*]
+indent_style = tab
```
ChrisSchreiber: if we are going to have a default tab policy I think spaces are safer

PR #226, opened 2020-01-23
Pls merge this parker

sjordan8:


modules/common/istio/istio-values.tpl
```diff
@@ -170,10 +170,10 @@ security:
   resources:
     requests:
       cpu: 20m
-      memory: 128Mi
+      memory: 32Mi
     limits:
-      cpu: 400m
-      memory: 256Mi
+      cpu: 1
```
mrparkers: Why are we increasing the limits so much here?  I think we should b focusing on bumping the requests, not the limits.

modules/common/istio/istio-values.tpl
```diff
@@ -115,24 +115,24 @@ tracing:
...
 prometheus:
   resources:
     requests:
       cpu: 200m
-      memory: 2Gi
+      memory: 1.5Gi
     limits:
-      cpu: 500m
-      memory: 4Gi
+      cpu: 800m
```
mrparkers: If prometheus was being throttled during the test, we should be increasing the requests, not the limits.

sjordan8: In my experience the request doesn't really effect throttling that much, in this case I was adjusting this prometheus pod's resource limits/requests to be equal to what it's set to in sandbox

PR #235, opened 2020-01-30
Eng 1417

jallen2112:
Harbor Integration. Allow pipelines in cluster to push images/charts to harbor or artifactory (Default to harbor). Map harbor/artifactory project credentials to dockercfg.


modules/lead/product/dockercfg.tf
```diff
@@ -0,0 +1,28 @@
...
+ 
+  vars = { 
+    email = "jenkins@liatr.io"
+    artifactory_url   = "https://artifactory.toolchain.${var.cluster_domain}/docker-registry/${var.product_name}"
+    artifactory_auth = base64encode(
+      "${data.kubernetes_secret.jenkins_artifactory_credential.data.username}:${data.kubernetes_secret.jenkins_artifactory_credential.data.password}",
+    )   
+    harbor_url = "https://harbor.toolchain.${var.cluster_domain}/${var.product_name}"
+    harbor_auth = base64encode(
+      "robot$imagepusher:${data.kubernetes_secret.product-harbor-creds.data.AUTH}",
```
JonRudy: Did the idea change of having multiple robot accounts specific for each product or repo?

jallen2112: No, harbor projects and auth resources are still created per product. The dockercfg contains both artifactory and harbor credentials so pipelines can push to either

JonRudy: hmm maybe i'm just not understanding the robot fully. I was assuming where you have `imagepusher` you would have to have the name of the specific robot account that gets created for that repo.

jallen2112: Every product has a robot account named `robot$imagepusher`, which is tied to the product by the specific auth token (robot$imagepusher:AUTHTOKEN). The token for the jordan-lab `robot$imagepusher` robot account won't work for the jon-lab `robot$imagepusher` and vice versa.

JonRudy: thanks for the explanation

modules/lead/product/toolchain.tf
```diff
@@ -24,6 +24,8 @@ data "template_file" "jenkins_values" {
     appDomain              = "apps.${var.cluster_domain}"
     builder_images_version = var.builder_images_version
     allow_anonymous_read   = var.enable_keycloak ? "false" : "true"
+    jenkins-repository-dockercfg = "jenkins-repository-dockercfg"
```
mrparkers: this should be referenced from `kubernetes_secret.jenkins_repository_dockercfg`, not hardcoded

PR #238, opened 2020-02-04
Update local dev cert issuer

ChrisSchreiber:
Probably not perfect but this was enough for me to be able to run terraform apply locally

### DIFF
--- a/stacks/environment-local/issuer.tf
+++ b/stacks/environment-local/issuer.tf
@@ -33,12 +33,10 @@ resource "kubernetes_cluster_role_binding" "cert_manager_cluster_role_binding" {
 module "staging_cluster_issuer" {
   source        = "../../modules/common/cert-issuer"
   namespace     = module.toolchain.namespace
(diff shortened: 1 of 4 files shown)


stacks/environment-local/issuer.tf
```diff
@@ -0,0 +1,46 @@
...
+    namespace = module.toolchain.namespace
+  }
+}
+
+module "staging_cluster_issuer" {
+  source        = "../../modules/common/cert-issuer"
+  namespace     = module.toolchain.namespace
+  issuer_name   = "staging-letsencrypt-http"
+  issuer_kind   = "ClusterIssuer"
+  issuer_type   = var.cert_issuer_type
```
mrparkers: I think you can just hardcode this to `selfsigned` and remove the `issuer_server` and `acme_solver` variables.

ChrisSchreiber: good call

PR #243, opened 2020-02-11
Harbor config

jallen2112:
Configuration to fully disable artifactory via feature flag. Pin kubernetes provider versions in product aws. Fix some harbor configuration.

### DIFF
--- a/modules/lead/product/main.tf
+++ b/modules/lead/product/main.tf
@@ -4,6 +4,7 @@ locals {
 
 provider "kubernetes" {
   alias = "toolchain"
+  load_config_file = "false"
 }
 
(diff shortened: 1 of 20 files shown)


modules/lead/product/main.tf
```diff
@@ -47,9 +50,10 @@ provider "helm" {
...
 }
 
-provider "helm" {
-  version = "0.10.4"
-  alias = "system"
-}
+#provider "helm" {
+#  version = "0.10.4"
+#  alias = "system"
+#}
```
mrparkers: Let's just delete this if it isn't needed instead of commenting it out

PR #249, opened 2020-02-27
Eng 1610

jallen2112:
Separate the creation of staging/production namespaces into `product-base` module. Rename `product` module to `product-jenkins`. Call `product-base` module from `product-jenkins`.


modules/lead/product-base/main.tf
```diff
@@ -0,0 +1,40 @@
...
+
+  override = [
+    "spec.template.metadata.annotations.sidecar\\.istio\\.io/inject=false"
+  ]
+}
+
+provider "kubernetes" {
+  alias = "system"
+  load_config_file = "false"
+}
```
mrparkers: looks like this is duplicated

modules/lead/product-base/production.tf
```diff
@@ -26,64 +26,6 @@ resource "helm_release" "production_product_init" {
   provider  = helm.production
 }
 
-resource "kubernetes_role" "jenkins_production_role" {
```
mrparkers: we actually need to keep this in the base module since we will always need some sort of role some CI (whether its Jenkins or codepipeline) to make changes.  So keep this here, but name the name more generic so it could apply to either Jenkins or CodePipeline

modules/lead/product-base/staging.tf
```diff
@@ -26,63 +26,6 @@ resource "helm_release" "staging_product_init" {
   provider  = helm.staging
 }
 
-resource "kubernetes_role" "jenkins_staging_role" {
```
mrparkers: same comment as above

modules/lead/product-base/outputs.tf
```diff
@@ -22,3 +14,10 @@ output "production_service_account" {
   value = module.production_namespace.tiller_service_account
 }
 
+output "ci_staging_role_name" {
+  value = module.staging_namespace.name
```
mrparkers: wrong output?

modules/lead/product-base/production.tf
```diff
@@ -151,3 +93,28 @@ resource "kubernetes_role_binding" "default_production_rolebinding" {
...
+    namespace = module.production_namespace.name
+
+    labels = {
+      "app.kubernetes.io/name"       = "ci"
+      "app.kubernetes.io/instance"   = "ci"
+      "app.kubernetes.io/managed-by" = "Terraform"
+    }
+
+    annotations = {
+      description = "Permission required for Continous Integration tools to get pods in production namespace"
```
mrparkers: I realize you didn't write this in this PR, but this isn't accurate anymore since this role allows CI to do pretty much everything. let's update this description

mrparkers:
1. let's remove all provider definitions from modules so they only exist within stacks.  you should still leave `provider` blocks that define aliases within modules, but nothing else.
2. the product-init module should be invoked within the product-base module.

modules/lead/product-base/main.tf
cplee: i think it's correct

PR #255, opened 2020-03-12
Eng 1613

jallen2112:
Feature flag `enable-aws-codestar`. Create necessary iam roles for aws ignite lab. Create S3 bucket, sqs queue and cloudwatch event rule for aws ignite lab.


mrparkers:
nitpick - to me, calling it `aws_codeservices` implies that it is a real service from AWS, as opposed to a collection of various AWS services that we're using together.  I think we could make this more clear if this was changed to `aws_code_services`.  thoughts?

jallen2112:
> nitpick - to me, calling it `aws_codeservices` implies that it is a real service from AWS, as opposed to a collection of various AWS services that we're using together. I think we could make this more clear if this was changed to `aws_code_services`. thoughts?

Yea I agree it makes it more clear, pretty easy fix I'll change it to `aws_code_services`

jallen2112:
> nitpick - to me, calling it `aws_codeservices` implies that it is a real service from AWS, as opposed to a collection of various AWS services that we're using together. I think we could make this more clear if this was changed to `aws_code_services`. thoughts?

Changed 👍

stacks/environment-aws/codeservices.tf
```diff
@@ -0,0 +1,275 @@
+resource "aws_s3_bucket" "code_services_bucket" {
```
ChrisSchreiber: The code services resources should go in a module so that they can be tested

modules/lead/product-aws/aws.tf
```diff
@@ -19,7 +19,7 @@ resource "aws_codebuild_project" "codebuild_build" {
 
   source {
     type            = var.source_type
-    location        = each.value.sourcelocation
+    location        = "https://git-codecommit.us-east-1.amazonaws.com/v1/repos/${each.value.repo}"
```
cplee: the source for codebuild should be codepipeline s3 bucket

modules/lead/product-aws/aws.tf
```diff
@@ -57,7 +57,7 @@ resource "aws_codebuild_project" "codebuild_staging" {
...
```
cplee: the source for codebuild should be codepipeline s3 bucket

modules/lead/product-aws/aws.tf
```diff
@@ -95,7 +95,7 @@ resource "aws_codebuild_project" "codebuild_production" {
...
```
cplee: the source for codebuild should be codepipeline s3 bucket

modules/lead/product-aws/aws.tf
jallen2112: The codebuild source seems to get overwritten by the source of codepipeline so this still works

jallen2112: Also that might create somewhat of a circular dependency in that the codebuild resource would need to wait for the execution of codepipeline before it had a valid source, but codepipeline can't run until the codebuild resource is created. Not sure if that would actually stop the creation of these resources though, I can try testing.

cplee: There isn't a circular dependency.  You just set the codebuild source.type to CODEPIPELINE and location to the S3 bucket.

jallen2112: Gotcha I thought you meant set the source.type to S3

PR #261, opened 2020-03-18
pass in corrrect vars to product-aws module from stack, fix codeservi…

jallen2112:
pass in corrrect vars to product-aws module from stack, fix codeservices bucketname

### DIFF
--- a/stacks/product-aws/variables.tf
+++ b/stacks/product-aws/variables.tf
@@ -19,3 +19,20 @@ variable "config_context" {
 variable "load_config_file" {
   default = false
 }
+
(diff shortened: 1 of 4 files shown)


stacks/product-aws/variables.tf
```diff
@@ -19,3 +19,20 @@ variable "config_context" {
...
+
+variable "pipelines" {
+  type = map(object({
+    type = string
+    repo = string
+    org = string
+  }))
+}
+
+variable "source_type" {
```
mrparkers: what is this variable for?  I don't see it being used

PR #264, opened 2020-03-19
Aws sqs role

sjordan8:


modules/environment/aws/code-services/main.tf
```diff
@@ -202,3 +202,47 @@ resource "aws_sqs_queue_policy" "code_services_queue_policy" {
...
+                "sts:*"
+            ],
+            "Effect": "Allow",
+            "Resource": "*"
+        },
+    ]
+}
+EOF
+
+  permissions_boundary = "arn:aws:iam::${data.aws_caller_identity.current.account_id}:policy/${aws_iam_policy.sqs_role_boundary.name}"
```
cplee: this shouldn't be a new policy..just use the developer policy

modules/environment/aws/code-services/main.tf
```diff
@@ -202,3 +202,47 @@ resource "aws_sqs_queue_policy" "code_services_queue_policy" {
...
+{
+    "Version": "2012-10-17",
+    "Statement": [
+        {
+            "Action": [
+                "sqs:*",
+                "sts:*"
+            ],
+            "Effect": "Allow",
+            "Resource": "*"
```
cplee: assume role policy shouldn't  have a resource.  also, the action should only be `sts:AssumeRole`.  also, needs a `Principal`

modules/environment/aws/code-services/main.tf
sjordan8: I didn't know there was a developer policy, just added it

modules/environment/aws/code-services/main.tf
```diff
@@ -202,3 +202,52 @@ resource "aws_sqs_queue_policy" "code_services_queue_policy" {
...
+  role   = aws_iam_role.sqs_role[0].name
+
+  policy = <<EOF
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+        "Action": [
+            "sqs:*",
+            "sts:*"
```
mrparkers: I think it's alright to dev with these permissions but I really don't think this should be merged until this is trimmed down.

modules/environment/aws/code-services/main.tf
```diff
@@ -202,3 +202,52 @@ resource "aws_sqs_queue_policy" "code_services_queue_policy" {
 }
 POLICY
 }
+
+resource "aws_iam_role" "sqs_role" {
```
mrparkers: this role / policy is for the event mapper, right?  if so, can you make these names a bit more specific?  this resource, its name, and the `aws_iam_role_policy` resource and its name should probably be changed

modules/environment/aws/code-services/main.tf
```diff
@@ -202,3 +202,52 @@ resource "aws_sqs_queue_policy" "code_services_queue_policy" {
...
+  "Statement": [
+    {
+      "Effect": "Allow",
+      "Principal": {
+        "Federated": "arn:aws:iam::${var.account_id}:oidc-provider/${replace(var.openid_connect_provider_url, "https://", "")}"
+      },
+      "Action": "sts:AssumeRoleWithWebIdentity",
+      "Condition": {
+        "StringEquals": {
+          "${replace(var.openid_connect_provider_url, "https://", "")}:aud": "sts.amazonaws.com"
```
mrparkers: I think it's fine to validate the audience but it's probably more important to validate the subject to ensure that only the service account for this event mapper can assume this role

PR #268, opened 2020-03-23
Added role for aws-event-mapper

sjordan8:


modules/environment/aws/code-services/main.tf
```diff
@@ -202,3 +202,52 @@ resource "aws_sqs_queue_policy" "code_services_queue_policy" {
...
+  "Statement": [
+    {
+      "Effect": "Allow",
+      "Principal": {
+        "Federated": "arn:aws:iam::${var.account_id}:oidc-provider/${replace(var.openid_connect_provider_url, "https://", "")}"
+      },
+      "Action": "sts:AssumeRoleWithWebIdentity",
+      "Condition": {
+        "StringEquals": {
+          "${replace(var.openid_connect_provider_url, "https://", "")}:aud": "sts.amazonaws.com"
```
mrparkers: while I think it's fine to validate the audience, we should absolutely also be validating the subject (`sub`) claim to ensure that only the aws event mapper service account gets these permissions.  take a look at how the other `aws_iam_role` resources within the aws stack are doing this.

modules/lead/sdm/operator-toolchain-values.tpl
```diff
@@ -22,6 +22,12 @@ product:
   rbac:
     serviceAccountAnnotations: ${product_service_account_annotations}
 
+aws-event-mapper:
+  enabled: ${aws_event_mapper_enabled}
```
mrparkers: the it looks like the variable you're passing to the `template_file` is `enable_aws_event_mapper`.  either this line or that variable will have to be changed.

modules/lead/sdm/operator-toolchain-values.tpl
```diff
@@ -22,6 +22,12 @@ product:
...
+  sqsUrl: ${sqs_url}
+  rbac:
+    serviceAccountAnnotations: ${aws_event_mapper_service_account_annotations}
```
mrparkers: I don't see this variable being passed to the `template_file` resource

modules/environment/aws/code-services/main.tf
sjordan8: I updated the trust policy to validate the subject claim

PR #269, opened 2020-03-23
Set SDM operator images using toolchain_image_repo var

ChrisSchreiber:


modules/lead/sdm/operator-toolchain-values.tpl
```diff
@@ -45,6 +51,8 @@ operators:
       value: ${region}
   jenkins:
     enabled: ${operator_jenkins_enabled}
+    image:
+      repository: ${image_repository}
```
mrparkers: How can these all share the same repository?  Does that mean the same docker image will be used for each of these?

mrparkers: Also the product operator image and converge image are missing

ChrisSchreiber: We could make the image repo the same value but I think long term we are going to break each of these into their own chart anyway.

ChrisSchreiber: Oops, I misread the first comment but everything should be fixed

PR #273, opened 2020-03-25
Eng 1716

jallen2112:
Feature flag for product type


modules/lead/sdm/variables.tf
```diff
@@ -38,11 +38,15 @@ variable "product_stack" {
   default = "product-local"
 }
 
-variable "product_vars" {
+variable "producttype_vars" {
```
mrparkers: nitpick: technically these are product variables, but we're using the `ProductType` to help define them.  would you mind renaming this variable back to `product_vars`?

jallen2112: Changed back to `product_vars`

PR #275, opened 2020-03-25
Eng 1720

ChrisSchreiber:
removes unnecessary helm providers and updates helm provider to 1.1.0


stacks/product-jenkins/main.tf
```diff
@@ -27,7 +27,7 @@ provider "kubernetes" {
 
 provider "helm" {
```
cplee: do we still need all 3 helm providers?

stacks/product-local/main.tf
```diff
@@ -35,7 +35,7 @@ provider "kubernetes" {
 }
 
 provider "helm" {
```
cplee: do we still need all 3 helm providers?

stacks/product-local/main.tf
```diff
@@ -9,7 +9,7 @@ provider "kubernetes" {
...
-  version         = "0.10.4"
+  version         = "1.1.0"
   alias           = "toolchain"
   namespace       = module.product_jenkins.toolchain_namespace
   tiller_image    = "gcr.io/kubernetes-helm/tiller:v2.15.1"
```
cplee: remove tiller_image

stacks/product-local/main.tf
ChrisSchreiber: Not the way we are using the products now. If we wanted to be able to have the products deploy to other clusters we would but we should have to add a bunch of other stuff anyway.

cplee: gotcha.  i'm fine keeping them around then

PR #276, opened 2020-03-26
Jcasc configmaps

tnishida1:
* Pulls out jcasc config into their own configmaps so we can manage them
* Creates a configmap to pass in the list of pipelines from the product
* Updates to local terraform so it somewhat works

Edit:
* Updated the pipeline configmap so now jcasc creates pipelines from the product list using the jobdsl plugin


modules/lead/product-jenkins/pipelines.tpl
```diff
@@ -0,0 +1,8 @@
+{
+%{ for name,pipeline in pipelines ~}
+  ${name}: {
+    type: ${pipeline.type},
+    repo: ${pipeline.repo},
+    org: ${pipeline.org}
+  },
```
cplee: i worry you will end up with invalid JSON due to the extra `,` here

tnishida1: for now I could do something like a substr to remove the extra `,`

cplee: did you test it?  if it works, then i'm good going forward with it as is

tnishida1: yup,  confirmed it works locally and just pushed up the change

stacks/environment-aws/main.tf
```diff
@@ -24,7 +24,7 @@ provider "kubernetes" {
 }
 
 provider "helm" {
-  version         = "1.1.0"
+  version         = "1.0.0"
```
ChrisSchreiber: did you mean to roll back this version?

tnishida1: was testing some changes, moved the version back to 1.1.0

mrparkers: I can approve this once that change is made, everything else LGTM

stacks/product-local/variables.tf
```diff
@@ -40,12 +29,27 @@ variable "enable_keycloak" {
   default = false
 }
 variable "builder_images_version" {
+  default = "v2.0.0"
```
ChrisSchreiber: not a major issue but I think setting default versions in our stacks creates problems more often than it really helps anything

tnishida1: gotcha, those were mostly there for testing so I removed this and the jenkins image version default

PR #281, opened 2020-04-01
Terraform fix for event-mapper role

sjordan8:


tests/aws/eks_test.go
```diff
@@ -1,196 +1,188 @@
 package aws
 
 import (
-	"fmt"
-	"os"
-	"flag"
-	"strings"
+    "flag"
```
mrparkers, line 4: is there a reason you're switching to using spaces instead of tabs?  golang generally prefers tabs and `go fmt` will undo all of these changes.

sjordan8: I messed the indentation up pretty bad. I forgot to run go fmt, I'll update that

PR #288, opened 2020-04-06
Eng 1745 && ENG-1746 && ENG-1747

jallen2112:
Make jenkins admin password creation sensitive. Pass cluster_domain as env var to operator-slack.


jallen2112:
Was able to get the fix for ENG-1747 in this PR as well. The `security-config.tpl` file which had the keycloak configuration, never made it into a configmap so just created a configmap to get picked up by jcasc.

PR #290, opened 2020-04-13
Update codeservices pipelines for lab

tnishida1:
* Adds permissions for codebuild to be able to pull/push images to ECR and deploy to EKS
* Updates to codebuild and codepipeline so they work with the labs


modules/environment/aws/eks/main.tf
```diff
@@ -32,8 +32,8 @@ EOF
 
   map_roles_extra = var.enable_aws_code_services ? [
     {
-      rolearn  = "arn:aws:iam::${data.aws_caller_identity.current.account_id}:role/aws-codebuild"
-      username = "user"
+      rolearn  = "arn:aws:iam::${data.aws_caller_identity.current.account_id}:role/codebuild-role-lead"
```
mrparkers, line 35: could we output this role arn from the `aws_iam_role. codebuild_role` resource in the code services module, instead of hardcoding it here?  or maybe pass the role name as a variable to both modules so its defined in one place?

modules/lead/product-aws/aws.tf
```diff
@@ -1,31 +1,46 @@
 resource "aws_codebuild_project" "codebuild_build" {
   for_each = var.pipelines
 
-  name          = "${var.product_name}-${each.value.repo}-build"
+  name          = "${each.value.repo}-build"
```
mrparkers, line 4: why did the product name get removed from the name?  is it okay to have two `aws_codebuild_project` resources with the same name?

modules/lead/product-aws/aws.tf
```diff
@@ -1,31 +1,46 @@
...
-  build_timeout = "5"
+  build_timeout = "20"
   service_role  = var.codebuild_role
 
   environment {
-    compute_type                = "BUILD_GENERAL1_SMALL"
-    image                       = "aws/codebuild/amazonlinux2-x86_64-standard:2.0"
+    privileged_mode             = true
+    compute_type                = "BUILD_GENERAL1_LARGE"
+    image                       = "489130170427.dkr.ecr.us-east-1.amazonaws.com/builder-image-skaffold:${var.builder_images_version}"
```
mrparkers, line 12: we're using a variable called `toolchain_image_repo` in lead-environments that could be used here so we aren't hardcoding the ecr repo

modules/lead/product-aws/aws.tf
```diff
@@ -1,31 +1,46 @@
...
     type                        = "LINUX_CONTAINER"
+    image_pull_credentials_type = "SERVICE_ROLE"
+
+    environment_variable {
+      name  = "SKAFFOLD_DEFAULT_REPO"
+      value = "774051255656.dkr.ecr.us-east-1.amazonaws.com/${var.product_name}"
```
mrparkers, line 18: won't this be different depending on the environment (sandbox vs prod)?  I think we used to have a variable in lead-environments called `product_image_repo` that we should probably use here to differentiate between environments.

modules/environment/aws/eks/main.tf
tnishida1: good call, I passed the output from the codeservices module into the eks module as a variable

modules/lead/product-aws/aws.tf
tnishida1: I originally removed it since sparky import was appending the product name to the repo name also so it was duplicating the name, I added it back since that should just be fixed in the SDM

modules/lead/product-aws/aws.tf
tnishida1: changed the value to be passed in as the toolchain_image_repo variable

modules/lead/product-aws/aws.tf
tnishida1: changed the value to be passed in from the product_image_repo variable and also passed it in as it's own separate env variable for the buildspec to use

modules/lead/product-aws/aws.tf
```diff
@@ -77,31 +116,51 @@ resource "aws_codebuild_project" "codebuild_staging" {
...
   environment {
-    compute_type                = "BUILD_GENERAL1_SMALL"
-    image                       = "aws/codebuild/amazonlinux2-x86_64-standard:2.0"
+    compute_type                = "BUILD_GENERAL1_LARGE"
+    image                       = "${var.toolchain_image_repo}/builder-image-skaffold:${var.builder_images_version}"
     type                        = "LINUX_CONTAINER"
+    image_pull_credentials_type = "SERVICE_ROLE"
+
+    environment_variable {
+      name  = "STAGING_NAMESPACE"
```
mrparkers, line 131: why do we need this for the production project?

modules/lead/product-aws/aws.tf
```diff
@@ -77,31 +116,51 @@ resource "aws_codebuild_project" "codebuild_staging" {
...
+      value = "${var.product_name}-staging"
+    }
+    environment_variable {
+      name  = "ISTIO_DOMAIN"
+      value = "${var.product_name}-staging.${var.cluster}.prod.liatr.io"
```
mrparkers, line 136: should this be `-production`?

stacks/product-aws/main.tf
```diff
@@ -45,6 +45,7 @@ provider "helm" {
 module "product-aws" {
   source                  = "../../modules/lead/product-aws"
   cluster_domain          = var.cluster_domain
+  cluster                 = split(".", var.cluster_domain)[0]
```
mrparkers, line 48: I'd rather just pass this in as a default product var than compute it like this

modules/lead/product-aws/aws.tf
tnishida1: should have been the production namespace env, changed it

modules/lead/product-aws/aws.tf
tnishida1: yeah, also changed

stacks/product-aws/main.tf
tnishida1: added it to the list of default product vars

modules/lead/product-aws/aws.tf
```diff
@@ -77,31 +116,51 @@ resource "aws_codebuild_project" "codebuild_staging" {
...
     type                        = "LINUX_CONTAINER"
+    image_pull_credentials_type = "SERVICE_ROLE"
+
+    environment_variable {
+      name  = "PROD_NAMESPACE"
+      value = "${var.product_name}-production"
+    }
+    environment_variable {
+      name  = "ISTIO_DOMAIN"
+      value = "${var.product_name}-production.${var.cluster}.prod.liatr.io"
```
mrparkers, line 136: should be `-prod` since that is the dns used for the wildcard cert.

mrparkers: we should be using `${var.cluster_domain}` instead of `${var.cluster}.prod.liatr.io` - otherwise this won't work in sandbox or any aws account we stand up for a new lab

modules/lead/product-aws/aws.tf
```diff
@@ -39,31 +58,51 @@ resource "aws_codebuild_project" "codebuild_build" {
...
     type                        = "LINUX_CONTAINER"
+    image_pull_credentials_type = "SERVICE_ROLE"
+
+    environment_variable {
+      name  = "STAGING_NAMESPACE"
+      value = "${var.product_name}-staging"
+    }
+    environment_variable {
+      name  = "ISTIO_DOMAIN"
+      value = "${var.product_name}-staging.${var.cluster}.prod.liatr.io"
```
mrparkers, line 78: we should be using `${var.cluster_domain}` instead of `${var.cluster}.prod.liatr.io` - otherwise this won't work in sandbox or any aws account we stand up for a new lab

mrparkers: might have to pass the cluster domain as a product variable too

modules/lead/product-aws/aws.tf
tnishida1: fixed to use `${var.cluster_domain}` and naming for pod

modules/lead/product-aws/aws.tf
tnishida1: fixed to use `${var.cluster_domain}`

PR #297, opened 2020-04-17
terraform configuration for google login via keycloak & keycloak test user

mrparkers:
when `enable_google_login` is true, google authentication will be added to the toolchain realm (used to be done in prod manually)

when `enable_test_user` is true, a test user named `casey` will be created in the toolchain realm (used to be done in sandbox manually)

### DIFF
--- a/stacks/environment-aws/lead.tf
+++ b/stacks/environment-aws/lead.tf
@@ -98,6 +98,11 @@ data "aws_ssm_parameter" "google_identity_provider_client_secret" {
   name  = "/${var.cluster}/google_identity_provider_client_secret"
 }
 
+data "aws_ssm_parameter" "test_user_password" {
+  count = var.enable_test_user ? 1 : 0
+  name  = "/${var.cluster}/test_user_password"
+}
+
 
 module "toolchain" {
   source                                 = "../../modules/lead/toolchain"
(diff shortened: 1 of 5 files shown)


stacks/environment-aws/lead.tf
```diff
@@ -88,31 +88,51 @@ data "aws_ssm_parameter" "prometheus_slack_webhook_url" {
...
+  image_whitelist                        = var.image_whitelist
+  elb_security_group_id                  = module.eks.aws_security_group_elb.id
+  artifactory_license                    = data.aws_ssm_parameter.artifactory_license.value
+  keycloak_admin_password                = data.aws_ssm_parameter.keycloak_admin_password.value
+  keycloak_postgres_password             = data.aws_ssm_parameter.keycloak_postgres_password.value
+  enable_google_login                    = var.enable_google_login
+  google_identity_provider_client_id     = var.enable_google_login ? data.aws_ssm_parameter.google_identity_provider_client_id[0].value : ""
+  google_identity_provider_client_secret = var.enable_google_login ? data.aws_ssm_parameter.google_identity_provider_client_secret[0].value : ""
+  enable_test_user                       = var.enable_test_user
+  test_user_password                     = var.enable_test_user ? data.aws_ssm_parameter.test_user_password[0].value : ""
```
ChrisSchreiber, line 121: do these params need to be added to the toolchain module in environment-local as well?

stacks/environment-aws/variables.tf
```diff
@@ -104,6 +104,14 @@ variable "enable_keycloak" {
...
 }
 
+variable "enable_google_login" {
+  default = false
+}
+
+variable "enable_test_user" {
+  default = false
+}
+
```
ChrisSchreiber, line 114: environment-local also?

stacks/environment-aws/lead.tf
mrparkers: yeah, but this already doesn't work anyways so I figured it would be better to wait until someone picks up the ticket to get local environment working again

stacks/environment-aws/variables.tf
mrparkers: I can go ahead and add this to environment-local, but I have no way of testing it since environment-local is already really out of date, and I don't want the scope of this PR to include fixing that as well.

stacks/environment-aws/lead.tf
ChrisSchreiber: we kind of keep kicking that can down the road and making it a bigger task for whoever takes it on

PR #298, opened 2020-04-17
give codebuild projects the ability to create network interfaces in vpc subnets

mrparkers:


stacks/environment-aws/codeservices.tf
```diff
@@ -1,3 +1,27 @@
...
+  }
+
+  filter {
+    name   = "cidr-block"
+    values = [
+      "*/18"
+    ]
+  }
+}
+
```
ChrisSchreiber, line 24: Can we just add the aws_subnet_ids data to the condservices module instead of passing the variable in?

PR #302, opened 2020-04-20
only changes to codebuild module

tnishida1:


modules/environment/aws/code-services/main.tf
```diff
@@ -1,3 +1,21 @@
+data "aws_subnet_ids" "eks_workers" {
```
mrparkers, line 1: you forgot to add the `data.aws_vpc.lead_vpc` here.  this won't run because line 2 points to a data source that doesn't exist

tnishida1: it was actually already in the file so I didn't need to add it
https://github.com/liatrio/lead-terraform/blob/deeafcc6b8e40d2cc204c377f617927c4ccff78e/modules/environment/aws/code-services/main.tf#L325-L329

mrparkers: my bad.  could you move this and the security group config to the top of the file so it's all in one spot?

tnishida1: sure, I've moved it up to the top

PR #311, opened 2020-04-27
Add GetCommit permission to event-mapper IAM role

sjordan8:


modules/environment/aws/code-services/main.tf
```diff
@@ -316,7 +316,14 @@ resource "aws_iam_policy" "event_mapper_role_policy" {
...
       "Resource": "*"
-    }
+    },
+	{
+	  "Action": [
+	    "codecommit:GetCommit"
+	  ],
+	  "Effect": "Allow",
+	  "Resource": "*"
+	}
```
mrparkers, line 326: can you fix the indentation so this json object aligns with the one before it?

sjordan8: My bad. Just updated it!

PR #321, opened 2020-05-01
update version and max autoscale

tnishida1:
Instead of increases the resources, I increased the max pods that istio can autoscale to. I assume 20 was what worked for ~30 participants, so I just increased it accordingly for 45


ChrisSchreiber:
I'm more concerned with pilot getting CPU throttled. I'm not sure scaling horizontally more will help. 

https://grafana.toolchain.lead.prod.liatr.io/d/xDjD87YWk/kubernetes-requests-and-limits-deployments?orgId=1&from=1588174964948&to=1588200976447&var-namespace=istio-system&var-deployment=istio-pilot&var-container=All&var-pod=All

tnishida1:
Then would we bump Pilot's CPU up to 11k mcpu or am I interpreting the graphs wrong? That seemed really high for just the CPU requests

mrparkers:
I think you're looking at the graph for the deployment as a whole, which is saying that all pods in the deployment were at one point using 10 CPUs.  If you scroll down further, you'll see the requests/limits/usage per pod, which shows most of them hitting 1 CPU at max.

I agree that I think it would be smarter to increase limits instead of the number of max replicas.

tnishida1:
ah okay that makes sense. I bumped the requests/limits up and reverted the max replicas back.

modules/common/istio/istio-values.tpl
```diff
@@ -40,10 +40,10 @@ global:
           memory: 10Mi
     resources:
       requests:
-        cpu: 100m
+        cpu: 1000m
         memory: 128Mi
       limits:
-        cpu: 1000m
+        cpu: 3000m
         memory: 512Mi
```
ChrisSchreiber, line 47: I think these are the default resources for the sidecar istio creates (not for the pilot container) but I could be wrong. Did you test this in sandbox to see what resources it changes?

PR #342, opened 2020-07-02
Added team_id variable in stack for environment_aws and environment_local

sjordan8:


mrparkers:
I might be missing something, but I don't see this value in any of the `terragrunt.hcl` files in lead-environments.  Can you please open a PR there to add this variable before this is merged?

sjordan8:
> I might be missing something, but I don't see this value in any of the `terragrunt.hcl` files in lead-environments. Can you please open a PR there to add this variable before this is merged?

team_id is secret variable, I'm assumed we didn't want that to be publicly available

mrparkers:
If it's a secret, then why are we expected to provide it while running lead-environments?  All of our other secrets are sourced from Vault, why is this one different?

PR #346, opened 2020-07-09
Vault secret refactor

sjordan8:


modules/tools/lab-partner/lab-partner-values.tpl
```diff
@@ -4,3 +4,9 @@ secrets:
   teamId: ${team_id}
 ingress:
   host: lab-partner.${namespace}.${cluster_domain}
+  annotations:
+    kubernetes.io/ingress.class: ${namespace}-nginx
+    nginx.ingress.kubernetes.io/force-ssl-redirect: "true"
+mongodb:
+  auth:
+    rootPassword: ${mongodb_password}
```
mrparkers, line 12: it looks like you're setting up lab partner to authenticate as the root user.  I think we should change this to use a regular user with permissions for that specific database.  we can use `auth.username` and `auth.password` to configure this.

modules/tools/lab-partner/lab-partner-values.tpl
```diff
@@ -4,3 +4,9 @@ secrets:
   teamId: ${team_id}
 ingress:
   host: lab-partner.${namespace}.${cluster_domain}
+  annotations:
+    kubernetes.io/ingress.class: ${namespace}-nginx
```
mrparkers, line 8: this ingress class should just be hardcoded to `toolchain-nginx`, or we should pass this in as a variable.  it isn't going to depend on the namespace that the lab partner is deployed to.

modules/tools/lab-partner/lab-partner-values.tpl
ChrisSchreiber: Is there a specific risk to using root for single purpose database? The problem is that if we use an unprivileged user we need to configure permissions to be able to access multiple databases.... botkit uses one for it's state and we need one for our data.

mrparkers: I'd prefer if we followed the principle of least privilege - so if the lab partner doesn't need to perform any administrative actions on the database, then we shouldn't use the root account.  similar to how I would avoid using the root account on linux boxes.

let's at least attempt to try this with a non-privileged account before we resort to using `root`.  I think we can accomplish this by using `auth.username` and `auth.password` to set up the account, and we can use [`initdbScripts` or `initdbScriptsConfigMap`](https://github.com/bitnami/charts/tree/master/bitnami/mongodb#initialize-a-fresh-instance) to provide initialization scripts that can create both of the required databases and allow this non privileged used to read and write to them.

if you're looking for a specific use case, perhaps we'd use Vault to issue temporary credentials to this bot similar to our plans to integrate Vault with the ignite lab.  in this case, Vault would use a privileged used to access this database, and the app would use a non privileged user, and we'd have to configure these `initdbScripts` regardless.

PR #349, opened 2020-07-24
Eng 1873

gesparza3:
This PR will enable the Product Operator to configure a role, policy, and backend connection for each instance of a Product that is stood up. Applications that wish to use these credentials will need to define Pod annotations to allow a vault-agent side car to be injected that will it's Product's roles/policies to obtain new credentials for it's MongoDB instance.

### DIFF
--- a/modules/lead/product-base/database.tf
+++ b/modules/lead/product-base/database.tf
@@ -6,20 +6,9 @@ module "database_namespace" {
     "appmesh.k8s.aws/sidecarInjectorWebhook" = "enabled"
   }
   annotations = {
-    name                                                     = "${var.product_name}-db"
-    "opa.lead.liatrio/ingress-whitelist"                     = "*.${var.product_name}-db.${var.cluster_domain}"
-    "opa.lead.liatrio/image-whitelist"                       = var.image_whitelist
-    "vault.hashicorp.com/agent-inject"                       = "true"
-    "vault.hashicorp.com/role"                               = "${var.product-name}-mongodb"
-    "vault.hashicorp.com/agent-inject-secret-mongodb.json"   = "database/creds/${var.product_name}-mongodb"
-    "vault.hashicorp.com/agent-inject-template-mongodb.json" = <<EOF
-  {{- with secret "database/creds/${var.product_name}-mongodb" -}}
-  {
-    "username": "{{ .Data.username }}",
-    "password": "{{ .Data.password }}"
-  }
-  {{- end }}
-EOF
+    name                                 = "${var.product_name}-db"
+    "opa.lead.liatrio/ingress-whitelist" = "*.${var.product_name}-db.${var.cluster_domain}"
+    "opa.lead.liatrio/image-whitelist"   = var.image_whitelist
   }
 
   providers = {
--- a/modules/lead/product-base/database.tf
+++ b/modules/lead/product-base/database.tf
@@ -6,11 +6,22 @@ module "database_namespace" {
     "appmesh.k8s.aws/sidecarInjectorWebhook" = "enabled"
   }
   annotations = {
-    name                                 = "${var.product_name}-db"
-    "opa.lead.liatrio/ingress-whitelist" = "*.${var.product_name}-db.${var.cluster_domain}"
-    "opa.lead.liatrio/image-whitelist"   = var.image_whitelist
+    name                                                     = "${var.product_name}-db"
+    "opa.lead.liatrio/ingress-whitelist"                     = "*.${var.product_name}-db.${var.cluster_domain}"
+    "opa.lead.liatrio/image-whitelist"                       = var.image_whitelist
+    "vault.hashicorp.com/agent-inject"                       = "true"
+    "vault.hashicorp.com/role"                               = "${var.product-name}-mongodb"
+    "vault.hashicorp.com/agent-inject-secret-mongodb.json"   = "database/creds/${var.product_name}-mongodb"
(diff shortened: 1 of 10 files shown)


mrparkers:
this still wip?

gesparza3:
> this still wip?

My bad, forgot to edit the title. This PR should be good-to-go, tested these changes in sandbox and verified that apps can use credentials from the Mongo instance in `<product-name>-db` namespace.

modules/lead/product-base/database.tf
```diff
@@ -36,5 +32,12 @@ resource "helm_release" "mongodb" {
...
-  values = [data.template_file.mongo_values.rendered]
-}
\ No newline at end of file
+  values = [
+    templatefile("${path.module}/mongo.tpl", {
+      mongodbRootPassword = random_password.mongodb_root_password.result
+    })
+  ]
+  depends_on = [
+    random_password.mongodb_root_password
```
mrparkers, line 41: nitpick: this is not needed since you are already referencing this resource on line 37 which creates an implicit dependency.

I don't care if you remove it or not, just wanted to let you know.

modules/lead/product-base/main.tf
```diff
@@ -21,3 +21,21 @@ provider "kubernetes" {
 provider "helm" {
   alias = "system"
 }
+
+data "kubernetes_secret" "vault_root_token_secret" {
+  provider = kubernetes.staging
+  metadata {
+    namespace = "toolchain"
+    name      = "vault-root-token"
```
mrparkers, line 29: maybe we should pass the `vault_root_token_secret` and `vault_namespace` as variables to the product so we don't have to hardcode these.  we can also use `vault_namespace` on line 34 as well

stacks/environment-aws/main.tf
```diff
@@ -51,3 +51,17 @@ provider "vault" {
...
 }
+
+data "kubernetes_secret" "vault_root_token_secret" {
+  metadata {
+    namespace = module.vault.vault_namespace
+    name      = module.vault.vault_root_token_secret
+  }
+}
+
+provider "vault" {
```
mrparkers, line 62: I don't have a great suggestion for you, but this is going to cause dependency issues when standing up the environment.  the provider initialization will fail on the first run since the vault instance its pointing to doesn't exist yet.

I'm wondering if now is the time to introduce a layered approach to the code in lead-terraform similar to how we handle the shared service cluster.

modules/lead/product-base/staging.tf
```diff
@@ -126,3 +130,75 @@ resource "kubernetes_role" "ci_staging_role" {
...
+
+resource "vault_database_secret_backend_role" "mongodb_role" {
+  backend = "mongodb"
+  creation_statements = [
+    jsonencode({
+      db = "admin"
+      roles = [
+        {
+          role = "readWrite"
+          db   = "database"
```
mrparkers, line 165: do we use a variable when configuring the database when standing up the mongodb instance?  if so, we should try to reuse that here.

modules/lead/product-base/staging.tf
```diff
@@ -126,3 +130,75 @@ resource "kubernetes_role" "ci_staging_role" {
...
+        }
+      ]
+    })
+  ]
+
+  db_name = vault_database_secret_backend_connection.mongodb.name
+  name    = local.vault_mongodb_role
+
+  default_ttl = 60
+  max_ttl     = 60
```
mrparkers, line 175: these ttl values were picked in order to demo this functionality, but in reality they should not be this aggressive.  perhaps every 5 or 10 minutes would be better

modules/lead/product-base/staging.tf
gesparza3: As far as I can tell we haven't actually started using these databases for market-summary so we aren't specifying which database we're using anywhere. I left it as `database` since we weren't using one yet and Mongodb [creates non-existent databases upon first use](https://docs.mongodb.com/manual/core/databases-and-collections/#create-a-database). 

If we want to go ahead and define a database here I can go about adding that configuration.

modules/lead/product-base/main.tf
gesparza3: Updated products to include variables `vault_root_token_name` and `vault_namespace` from `vault` module outputs.

modules/lead/product-base/database.tf
gesparza3: Good eye here, I initially added this to make _sure_ the resource was there for the helm chart installation but as long as Terraform knows to wait I'm fine with removing it.

stacks/environment-aws/main.tf
gesparza3: Yikes, I can see now how this is a problem...will bring it up in standup to get more feedback and thoughts.

modules/lead/product-base/staging.tf
gesparza3: Increased vault token timeout to 10 minutes

modules/lead/product-base/staging.tf
gesparza3: Updated database value used in this secret backend role to match the database used in market-summary:
- [Connection string used by market-summary](https://github.com/liatrio/springtrader-marketsummary/blob/98edd158c2d155106be68001f72d0859db138d3d/src/repository/index.js#L16)
- [NODE_ENV value](https://github.com/liatrio/springtrader-marketsummary/blob/f5af8c0b8052a7c82814724cdf16f2baa430c644/Jenkinsfile#L27)

mrparkers:
Closed in favor of #406.

PR #350, opened 2020-08-03
Eng 1879

gesparza3:
This PR will install Rode as a part of the LEAD Toolchain. Installation is handled through use of a feature flag. In order to install Rode the following resources need to be created:
- Certificates for both Rode and it's corresponding Grafeas instance
- AWS IAM Role and Policy to allow the ECR collector to utilize SQS queues

As a part of installing Rode onto the LEAD Toolchain, this PR also removes old Grafeas/Kritis config since Rode will be replacing their previous functionalities.


modules/lead/toolchain/rode.tf
```diff
@@ -0,0 +1,63 @@
...
+}
+
+data "template_file" "rode_values" {
+  template = file("${path.module}/rode-values.tpl")
+
+  vars = {
+    iam_arn      = var.rode_service_account_arn
+    grafeas_cert = module.grafeas_certificate.cert_name
+    rode_cert    = module.rode_certificate.cert_name
+    ingress_hostname     = "rode.${module.toolchain_namespace.name}.${var.cluster}.${var.root_zone_name}
```
mrparkers, line 47: missing an end quote here?

modules/lead/toolchain/rode.tf
```diff
@@ -0,0 +1,63 @@
...
+  certificate_crd = var.crd_waiter
+  wait_for_cert   = true
+}
+
+data "helm_repository" "liatrio-harbor" {
+  name = "liatrio-harbor"
+  url  =  "https://harbor.toolchain.lead.prod.liatr.io/chartrepo/public"
+}
+
+data "template_file" "rode_values" {
```
mrparkers, line 40: I think the `template_file` data source is deprecated now.  let's convert this to using the native [`templatefile`](https://www.terraform.io/docs/configuration/functions/templatefile.html) function instead.

modules/lead/toolchain/rode.tf
```diff
@@ -0,0 +1,63 @@
...
+    iam_arn      = var.rode_service_account_arn
+    grafeas_cert = module.grafeas_certificate.cert_name
+    rode_cert    = module.rode_certificate.cert_name
+    ingress_hostname     = "rode.${module.toolchain_namespace.name}.${var.cluster}.${var.root_zone_name}
+  }
+}
+
+resource "helm_release" "rode" {
+  count      = var.enable_rode ? 1 : 0
+  repository = data.helm_repository.liatrio-harbor.metadata[0].name
```
mrparkers, line 53: why is `.metadata[0]` needed here?  that data source doesn't have a `metadata` argument

modules/lead/toolchain/rode-values.tpl
```diff
@@ -0,0 +1,30 @@
...
+rbac:
+  serviceAccountAnnotations:
+    eks.amazonaws.com/role-arn: ${iam_arn}
+
+localstack:
+  enabled: false
+
+resources:
+  limits:
+    cpu: 1
```
mrparkers, line 17: this is pretty steep, what made you decide on using 1 cpu here?

modules/lead/toolchain/rode.tf
```diff
@@ -0,0 +1,63 @@
...
+  enabled         = var.enable_rode
+  name            = "grafeas-cert"
+  namespace       = var.namespace
+  domain          = "grafeas-server"
+  issuer_name     = module.ca-issuer.name
+  certificate_crd = var.crd_waiter
+  wait_for_cert   = true
+}
+
+module "rode_certificate" {
```
mrparkers, line 23: I'm a bit confused by why this is needed.  the rode chart doesn't currently support passing your own certificates to it.  all you can do is provide a name for the secret that the rode chart creates for you. (https://github.com/liatrio/rode/blob/master/helm-chart/rode/templates/secrets.yaml#L29).

to be clear, I think rode definitely should support passing your own certificates (and potentially having that be the _only_ option), but since that's not how the chart works now, maybe this should be removed.

modules/lead/toolchain/rode.tf
```diff
@@ -0,0 +1,63 @@
+module "ca-issuer" {
```
mrparkers, line 1: tf style nitpick: names of resources and modules should use underscores instead of dashes.

modules/lead/toolchain/rode.tf
gesparza3: Thank you much, I think I accidentally deleted this will fixing some formatting

modules/lead/toolchain/rode.tf
gesparza3: Other examples I found(like here in [artifactory.tf](https://github.com/liatrio/lead-terraform/blob/22b6af99946d1c689550973610a4f2547a5e11d1/modules/lead/toolchain/artifactory.tf#L126)) used this attribute which according to the documentation is exported to be the name of the helm repository: https://registry.terraform.io/providers/hashicorp/helm/latest/docs/data-sources/repository#name.

I'm fine with updating this to `data.helm_repository.liatrio_harbor.name` to make it more clear, just want to keep it consistent with others I found.

modules/lead/toolchain/rode-values.tpl
gesparza3: The `1 cpu` is actually the default limit in the Rode Helm chart, to be clear I was running into `OOM Killed` problems when testing in sandbox so I included the resources here to bump up the memory.

modules/lead/toolchain/rode.tf
gesparza3: Ah I see that now, that was just a failure of understanding the functionality on my part. I've made updated changes to pass a secret name for both Rode and Grafeas and removed the certificate creation.

modules/lead/toolchain/rode-values.tpl
mrparkers: We should probably change that in the rode chart then, but I think 1 CPU is definitely too hard.

Just so you understand, CPU usage doesn't have any bearing on whether or not a pod gets `OOM Killed`.  Since `OOM` stands for "out of memory".

gesparza3: No problem with decreasing the cpu here, and I understood that the issue here was just with the memory; I just thought it was worth defining all the resource values we're using (_including the defaults from Rode_) for some more visibility.

I went ahead and did some more testing in sandbox to come up with more reserved values that performed well for me. I imagine these will adjust as we use proceed with dogfooding within our clusters.

modules/lead/toolchain/rode.tf
mrparkers: I guess I don't care that much since this data source is deprecated anyways and we'll be removing it when upgrading the helm provider.

PR #353, opened 2020-08-24
Eng 1901

gesparza3:
This PR removes configuration for Mailhog as well as references to Mailhog/SMTP values.


stacks/environment-local/lead.tf
```diff
@@ -48,13 +47,6 @@ module "toolchain" {
 
   harbor_registry_disk_size    = "200Gi"
   harbor_chartmuseum_disk_size = "100Gi"
-
-
-  smtp_host       = "mailhog"
-  smtp_port       = "1025"
-  smtp_username   = ""
-  smtp_password   = ""
-  smtp_from_email = "noreply@liatr.io"
```
mrparkers, line 57: These should also be removed from the toolchain module's `variables.tf` if we aren't going to pass them anymore.

mrparkers:
Can you double check that `terragrunt plan` works against sandbox with these changes?

gesparza3:
Just ran against sandbox and verified it works 👍

PR #354, opened 2020-08-24
Eng 1895

gesparza3:
This PR removes Prometheus config from the toolchain module and into it's own module under the `modules/tools` folder. The AWS stack and local stack have been updated to reference this new structure.


stacks/environment-local/lead.tf
```diff
@@ -111,3 +109,11 @@ module "lab_partner" {
   lab_partner_version         = var.lab_partner_version
 }
 
+module "prometheus-operator" {
+  source = "../../modules/tools/prometheus"
```
mrparkers, line 113: did `terragrunt plan` not fail here?  I'd expect it to since it has moved locations.

can you fix this and verify that `terragrunt plan` works with these changes?

gesparza3: Yup, just tested this branch against sandbox. Caught this issue and pushed up the fix

PR #356, opened 2020-08-26
Eng 1897 - Create SonarQube Module

jonathan-dorsey:
Changes to pull sonarqube out of the toolchain and into its own module. Most of the code was moved as-is, but changes were made to point to a new helm repo for an updated version of the chart and to add a new terratest case for sonarqube. I also commented out a resource that wasn't being used so we can look at whether we should re-introduce it or remove it completely.


jonathan-dorsey:
@gesparza3 @ChrisSchreiber - merge conflicts were resolved by @gesparza3  and I cleaned up one additional problem found from that work. Plan appears to be correct, but we may want to decide what to do with sonar_jenkins_password. I commented it out for now since it's unused, but we may want to remove it altogether if there is no plan for using it.

modules/tools/sonarqube/main.tf
```diff
@@ -3,22 +3,27 @@ resource "random_string" "sonarqube_db_password" {
...
 }
 
-resource "random_string" "sonar_jenkins_password" {
-  length  = 10
-  special = false
-}
+#resource "random_string" "sonar_jenkins_password" {
+#  length  = 10
+#  special = false
+#}
```
mrparkers, line 9: let's remove this if it's no longer needed

jonathan-dorsey: sounds good - resource removed

PR #359, opened 2020-08-27
Updated harbor and keycloak provider config to point to Terraform registry

gesparza3:
This PR updates the required_providers config to point to the newly published `harbor` and `keycloak` providers.

### DIFF
--- a/stacks/environment-aws/lead.tf
+++ b/stacks/environment-aws/lead.tf
@@ -33,6 +33,7 @@ module "toolchain" {
   google_identity_provider_client_secret = var.enable_google_login ? data.vault_generic_secret.keycloak.data["google-idp-client-secret"] : ""
   enable_test_user                       = var.enable_test_user
   test_user_password                     = var.enable_test_user ? data.vault_generic_secret.keycloak.data["test-user-password"] : ""
+  harbor_admin_password                  = random_string.harbor_admin_password.result
   enable_istio                           = var.enable_istio
   enable_artifactory                     = var.enable_artifactory
   enable_gitlab                          = var.enable_gitlab
--- a/stacks/environment-aws/lead.tf
+++ b/stacks/environment-aws/lead.tf
@@ -10,6 +10,10 @@ data "vault_generic_secret" "keycloak" {
   path = "lead/aws/${data.aws_caller_identity.current.account_id}/keycloak"
 }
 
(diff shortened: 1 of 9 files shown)


mrparkers:
can you help me understand the context behind the "tf upgrade todo" comments?

gesparza3:
> can you help me understand the context behind the "tf upgrade todo" comments?

This was generated from running `terrafrom 0.13upgrade`, I've pushed up a change that cleaned up those comments.

gesparza3:
This should now be good to review, I've gotten the plan working after moving providers over to `environment-aws`. The one change I don't like is moving the `harbor_admin_password` generation to `environment-aws` but this was necessary as I needed the provider to be configured with the password prior to instantiating the Harbor module now that it's been pulled out of toolchain. This should be able to be resolved as a part of creating the Harbor modules, since we can set it as an output variable so it can be set on the Harbor provider before applying the configuration.

mrparkers:
> The one change I don't like is moving the harbor_admin_password generation to environment-aws but this was necessary as I needed the provider to be configured with the password prior to instantiating the Harbor module now that it's been pulled out of toolchain.

What do you think about generating this password one time and storing it in Vault?  Then we can pull it in with the `vault_generic_secret` data source.

gesparza3:
That sounds fine to me, we do it in other places like for the `keycloak_admin_password` so it makes sense to do it here too.

gesparza3:
Updated to retrieve `harbor_admin_password` from Vault.

stacks/environment-aws/lead.tf
```diff
@@ -33,6 +37,7 @@ module "toolchain" {
   google_identity_provider_client_secret = var.enable_google_login ? data.vault_generic_secret.keycloak.data["google-idp-client-secret"] : ""
   enable_test_user                       = var.enable_test_user
   test_user_password                     = var.enable_test_user ? data.vault_generic_secret.keycloak.data["test-user-password"] : ""
+  harbor_admin_password                  = data.vault_generic_secret.keycloak.data["admin-password"]
```
mrparkers, line 40:
```suggestion
  harbor_admin_password                  = data.vault_generic_secret.harbor.data["admin-password"]
```

gesparza3:
Made that fix. I don't know how I missed that, sorry about that.

PR #374, opened 2020-09-01
Eng 1906

gesparza3:
This PR reorganizes `lead-terraform` to use a umbrella module which groups applications/tools by referencing submodules. These groups are referred to as `stages` and are intended to be referenced as a stack of stages using Terragrunt. The previously used `environment-aws` stack has been decomposed into these stages under `stages/<layer>/lead`. Additionally this PR also moves over the configuration for `sharedsvc` into this new directory structure. Previously the entire stack under `environment-aws` shared a single `variables.tf`; this PR has moved these variables to their appropriate stages and created/deprecated variables to support this new approach for layering resources.

Once this PR is merged we can begin standing up new environments using these new stages.


stages/apps/lead/elasticstack.tf
```diff
@@ -1,40 +1,38 @@
...
-  source = "../../modules/tools/elasticsearch"
+  source = "../../../modules/tools/elasticsearch"
 
-  cert_manager_crd_waiter = module.cert_manager.crd_waiter
   namespace               = module.elasticsearch_namespace.name
   root_zone_name          = var.root_zone_name
   disk_size               = "50Gi"
+
+  depends_on = [
+    module.cert_manager
```
mrparkers, line 19: this looks so much nicer.

PR #377, opened 2020-09-02
Fixes #376 added output for random_password

### LINKED ISSUE #376: Add output for random_password
Add output for random_password

AP

Output updated for provider "random"

jgetzie:
Adding output for random_password

### DIFF
--- a/stacks/environment-azure/outputs.tf
+++ b/stacks/environment-azure/outputs.tf
@@ -1,0 +1,3 @@
+output "password" {
+    value = random_password.password.result
+}


stacks/environment-azure/outputs.tf
```diff
@@ -0,0 +1,3 @@
+output "password" {
+    value = random_password.password.result
```
mrparkers, line 2: I'm not entirely sure I understand what this is for, but typically we'd add `sensitive = true` so this value doesn't get printed in your shell in plaintext when running `terragrunt apply`.

stacks/environment-azure/outputs.tf
```diff
@@ -0,0 +1,3 @@
...
+}
```
ChrisSchreiber, line 3: Probably a good idea to mark the output as sensitive. It will obscure the value in the terraform output.

stacks/environment-azure/outputs.tf
jgetzie: Under review with team

jgetzie:
Not needed, abandoning PR

PR #378, opened 2020-09-02
Removing modules/lead/infrastructure from source

jonathan-dorsey:
Removed the directory from source and corresponding module in the environment-local lead.tf. The environment-local will be broken until we can refactor it to use the new stages being introduced by ENG-1906


jonathan-dorsey:
FYI - Had an incidental change on original commit. Added a new commit to remove that change, squashed, and then force-pushed to remove it from the PR to avoid confusion.

gesparza3 (approved):
LGTM! We have a ticket in the backlog(ENG-1652) which we can pull in to address the broken local environment.

PR #394, opened 2020-09-03
ENG-1912-modules+k8s1.16

ChrisSchreiber:
Update several module version to work with k8s 1.16
Fix misc test problems from recent module changes


gesparza3 (approved):
Looks good to me! Nice to get these updated

PR #408, opened 2020-10-06
increase success threshold for mongodb liveness and readiness probes

mrparkers:
without this change, there's a small chance that the mongodb pod is not actually ready for authentication attempts by the time that the helm chart finishes deploying, causing an issue with the terraform code that sets up a connection between vault and mongodb.


alexashley (approved):

mrparkers:
turns out the success threshold for the liveness probe has to be 1.  so I removed that.

PR #411, opened 2020-10-14
add essential node tolerations to jenkins workers

mrparkers:
also removed it from kube-janitor and kube-downscaler since it wasn't being used anyways


ChrisSchreiber (approved):
Looks good

PR #420, opened 2021-01-27
Module changes to support k8s 1.16

ChrisSchreiber:


modules/tools/aws-node-termination-handler/main.tf
```diff
@@ -0,0 +1,13 @@
...
+  chart      = "aws-node-termination-handler"
+  version    = "0.13.3"
+  namespace  = var.namespace
+  name       = "aws-node-termination-handler"
+  timeout    = 600
+
+  values     = [
+    file("${path.module}/values.yaml")
+  ]
+}
```
mrparkers, line 13: is this our first time using `aws-node-termination-handler`?  before we were using something else like `k8s-spot-termination-handler`.  I'm sure this works fine, but was this applied to sandbox yesterday?  so we should be able to know how that worked last night.  any info on this?

modules/tools/cluster-autoscaler/charts/Chart.yaml
```diff
@@ -0,0 +1,20 @@
...
+    name: mgoodness
+  - email: guyjtempleton@googlemail.com
+    name: gjtempleton
+  - email: scott.crooks@gmail.com
+    name: sc250024
+name: cluster-autoscaler
+sources:
+  - https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler
+type: application
+version: 9.4.0
```
mrparkers, line 20: do we need to vendor this chart?  looks like its available here: https://artifacthub.io/packages/helm/cluster-autoscaler/cluster-autoscaler-chart

ChrisSchreiber: The artifactory hub is a copy from the deprecated repo and I'm not sure it is being actively maintained. The cluster-autoscaler repo links to the [official chart](https://github.com/kubernetes/autoscaler/tree/master/charts) in the repo which is more up to date.

modules/tools/aws-node-termination-handler/main.tf
ChrisSchreiber: It deployed fine in sandbox and all of the logs look fine. I I don't really have a good way to test one of the spot instances being removed.

The new one also has a prometheus metrics endpoint which would be cool to look at the data there and maybe setup alerts but it's not really a priority right now.

modules/tools/cluster-autoscaler/charts/Chart.yaml
ChrisSchreiber: Took another look and it is being maintained but there are two different charts https://artifacthub.io/packages/helm/cluster-autoscaler/cluster-autoscaler. 

Updated module to use published chart.

mrparkers (approved):

PR #421, opened 2021-01-27
Updating Jenkins provider 

BFisch14:
- Config context default to null
- Pinning version of Kubernetes to 2.0.1
- Updated helm chart repo for Jenkins

### DIFF
--- a/stacks/product-jenkins/versions.tf
+++ b/stacks/product-jenkins/versions.tf
@@ -1,4 +1,10 @@
 
 terraform {
   required_version = ">= 0.12"
+  required_providers {
+    kubernetes = {
+      source = "hashicorp/kubernetes"
+      version = "= 1.13.3"
+    }
+  }
 }
--- a/stacks/product-jenkins/versions.tf
+++ b/stacks/product-jenkins/versions.tf
@@ -3,8 +3,12 @@ terraform {
   required_version = ">= 0.12"
   required_providers {
     kubernetes = {
-      source = "hashicorp/kubernetes"
+      source  = "hashicorp/kubernetes"
       version = "= 2.0.1"
+    }
+    helm = {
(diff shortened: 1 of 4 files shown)


stacks/product-jenkins/versions.tf
```diff
@@ -1,4 +1,10 @@
 
 terraform {
   required_version = ">= 0.12"
+  required_providers {
```
mrparkers, line 4: I am not sure if this will work.  this is a terraform 0.13 feature, and the product operator that consumes this terraform still uses 0.12

BFisch14: Changed it to >= 0.13

ChrisSchreiber (approved):

stacks/product-jenkins/versions.tf
mrparkers: that won't actually solve the problem, it'll actually make it worse - the container the product operator uses doesn't have that version of tf installed

mrparkers: https://www.terraform.io/docs/language/providers/requirements.html - `required_providers` was added in terraform 0.13.  so if we're still using 0.12, I would expect this to error out immediately.

mrparkers (changes requested):
requesting changes to prevent a merge because I'm confident that this won't work until we update the product operator to use tf 0.13: https://github.com/liatrio/lead-sdm-operators/blob/master/operator-product/converge-image/Dockerfile#L9

stacks/product-jenkins/versions.tf
mrparkers: let's just remove `required_providers` entirely since it is dead code anyways.  we can add it when we update to 0.13

mrparkers (approved):

PR #425, opened 2021-02-18
prevent vault root token change

ChrisSchreiber:
this should prevent cases where the vault root token stored in a k8s secret is replaced with an empty string because the shell script resource is recreated, tries to unlock vault again and returns an error.


mrparkers (approved):
I wish I knew why the shell resource kept trying to do this in the first place, but this works in the meantime

PR #426, opened 2021-02-19
docker registry mirror

ChrisSchreiber:
configure nodes docker daemon to use registry mirror via ec2 instance user data

### DIFF
--- a/modules/environment/aws/eks/main.tf
+++ b/modules/environment/aws/eks/main.tf
@@ -2,10 +2,14 @@ data "aws_caller_identity" "current" {
 }
 
 locals {
-  ssm_init = <<EOF
+  userdata = <<EOF
 yum install -y amazon-ssm-agent
 systemctl start amazon-ssm-agent
 systemctl enable amazon-ssm-agent
+
+echo '{"registry-mirrors": [${ var.docker_registry_mirror != "" ? format("\"%s\"", var.docker_registry_mirror) : ""}]}' | cat /etc/docker/daemon.json - | jq -s '.[0] * .[1]' > /tmp/daemon.json
+mv /tmp/daemon.json /etc/docker/daemon.json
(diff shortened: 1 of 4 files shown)


modules/environment/aws/eks/main.tf
```diff
@@ -2,10 +2,14 @@ data "aws_caller_identity" "current" {
 }
 
 locals {
-  ssm_init = <<EOF
+  userdata = <<EOF
 yum install -y amazon-ssm-agent
 systemctl start amazon-ssm-agent
 systemctl enable amazon-ssm-agent
+
+echo '{"registry-mirrors": [${ var.docker_registry_mirror != "" ? format("\"%s\"", var.docker_registry_mirror) : ""}]}' | cat /etc/docker/daemon.json - | jq -s '.[0] * .[1]' > /tmp/daemon.json
```
mrparkers, line 10: what happens when `registry-mirrors` is set to an empty string?

ChrisSchreiber: it sets no registry-mirrors `registry-mirrors: []` if we don't handle that case we end up with `registry-mirrors: [ "" ]` which breaks the daemon

mrparkers: maybe I'm misreading this, but it seems like `${ var.docker_registry_mirror != "" ? format("\"%s\"", var.docker_registry_mirror) : ""}` will evaluate to `""` if `var.docker_registry_mirror` is set to `""`.

ChrisSchreiber: yeah, it is hard to read but I did test both cases. if the string is not empty it wraps the string value in quotes "\"http://hostname\"" evaluates to "http://hostname" otherwise "" evaluates to nothing.

mrparkers (approved):

PR #427, opened 2021-02-19
migrate to new stable chart location

mrparkers:
https://helm.sh/blog/new-location-stable-incubator-charts/

I've already done the state surgery in sandbox.  haven't done it in prod yet.  we'll want to do that before any further applies so releases don't get removed and recreated


ChrisSchreiber (approved):
I'm going to add jira tickets to update each of these since this repo is also deprecated but this is good for now.

mrparkers:
yeah this will at least allow us to run `terragrunt apply` lol

PR #428, opened 2021-02-19
install monitoring stack to shared-svc cluster

mrparkers:
https://grafana.internal.services.liatr.io/


modules/tools/docker-registry/values.yaml.tpl
```diff
@@ -36,3 +38,10 @@ configData:
       threshold: 3
   proxy:
     remoteurl: https://registry-1.docker.io
+resources:
+  requests:
+    cpu: 100m
+    memory: 128Mi
+  limits:
+    cpu: 200m
+    memory: 256Mi
```
mrparkers, line 47: just setting some sane defaults here because the defaults provided by the namespace module definitely won't be enough.  we can tweak this after load testing

ChrisSchreiber (approved):

PR #432, opened 2021-02-25
Changed harbor version to 2.0.6, removed clair, and added trivy

PaulDHenson:


mrparkers (commented):
any reason in particular why we're not just using the latest version `v2.2.0`?

modules/tools/harbor/main.tf
```diff
@@ -122,7 +122,7 @@ resource "helm_release" "harbor" {
   name       = "harbor"
   namespace  = var.namespace
   chart      = "harbor"
-  version    = "1.3.0"
+  version    = "2.0.6"
```
mrparkers, line 125: I don't see this version in the helm chart releases, are you sure this is correct?  https://github.com/goharbor/harbor-helm/releases

PaulDHenson: Oh it looks like I'm referencing the OSS version, and not the release version. I'll change that now.

PR #435, opened 2021-03-05
Terraform 14 upgrade

PaulDHenson:
Updated the version of Terraform that product-jenkins was using from 12 to 13 and then to 14, expanded the IAM role permissions for the S3 bucket to compensate for the update. Listed the providers in required_providers section inside the versions.tf file as required by the Terraform 13 update.

### DIFF
--- a/modules/lead/product-jenkins/versions.tf
+++ b/modules/lead/product-jenkins/versions.tf
@@ -10,16 +10,8 @@ terraform {
       source = "hashicorp/helm"
     }
     keycloak = {
-      # TF-UPGRADE-TODO
-      #
-      # No source detected for this provider. You must add a source address
-      # in the following format:
-      #
-      # source = "your-registry.example.com/organization/keycloak"
-      #
-      # For more information, see the provider source documentation:
-      #
-      # https://www.terraform.io/docs/configuration/providers.html#provider-source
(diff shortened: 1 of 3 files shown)


modules/lead/product-jenkins/versions.tf
```diff
@@ -1,10 +1,26 @@
...
-      source = "liatrio/harbor"
+      source  = "liatrio/harbor"
       version = "= 0.3.3"
     }
+    helm = {
+      source = "hashicorp/helm"
+    }
+    keycloak = {
+      source = "mrparkers/keycloak"
+      version = "= 2.0.0-rc.0"
```
mrparkers, line 14:
```suggestion
      version = "= 2.3.0"
```

stacks/product-jenkins/versions.tf
```diff
@@ -10,5 +10,19 @@ terraform {
...
     }
+
+    harbor = {
+      source  = "liatrio/harbor"
+      version = "= 0.3.3"
+    }
+
+    keycloak = {
+      source  = "mrparkers/keycloak"
+      version = "= 2.0.0-rc.0"
```
mrparkers, line 21:
```suggestion
      version = "= 2.3.0"
```

mrparkers (approved):

PR #445, opened 2021-03-22
Ray 24 add jcr

gesparza3:
This PR creates adds modules for creating an instance of Artifactory-JCR for publishing OCI compliant artifacts. The included modules will create an instance of JCR in the `apps` stage and create Docker & Helm repos within the `config` stage. Each new instantiation of a product will create a new _user_ in Jenkins to allow pipelines to publish to Artifactory.


RobertKelly (approved):
Looks good!

stages/apps/lead/variables.tf
```diff
@@ -75,6 +75,10 @@ variable "enable_harbor" {
   default = true
 }
 
+variable "enable_artifactory_jcr" {
+  default = true
```
mrparkers, line 79: I don't think this should be `true`.  lead-terraform's primary use case is still for our lab purposes, so we don't want to accidentally leave this on for a lab.  I'd prefer this be `false` and we can explicitly override this in lead-environments.

modules/lead/product-jenkins/artifactory.tf
```diff
@@ -1,34 +1,12 @@
...
-  data = {
-    "settings.xml" = data.template_file.maven_settings[0].rendered
-  }
-
-  type = "Opaque"
+resource "artifactory_user" "robot" {
+  count      = var.enable_artifactory ? 1 : 0
+  name       = "robot-${var.product_name}" 
+  email      = "robot-${var.product_name}@liatrio.com"
+  password   = data.random_password.robot_password.result
```
mrparkers, line 11: is there a reason these changes are within the `artifactory.tf` file and not a new file?  is the plan to completely deprecate artifactory usage within lead-terraform?

mrparkers (changes requested):

modules/lead/product-jenkins/artifactory.tf
RobertKelly: Maybe?

mrparkers: if so, I'd still prefer changes like that to show up in another PR, and not do some of them in this one, then follow up with the rest.

can we add these changes in a new file?  also, how do these user credentials get seeded to jenkins?

gesparza3: Artifactory already is deprecated or it was back when we did the big refactor. This was a leftover file that I repurposed.

stages/apps/lead/artifactory.tf
```diff
@@ -0,0 +1,12 @@
+data "vault_generic_secret" "artifactory" {
+  path = "lead/aws/${data.aws_caller_identity.current.account_id}/artifactory"
```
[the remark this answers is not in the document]
gesparza3: No not yet, I don't have access on my end to do that. Figured we'd resolve any comments on this PR running the terraform against our cluster.

modules/lead/product-jenkins/artifactory.tf
mrparkers: gotcha, can we rename the file then so it's clear that this is specifically for JCR?

stages/apps/lead/artifactory.tf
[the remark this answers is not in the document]
mrparkers: @dogbern should have access.  I'll open a PR now to get you access too

stages/config/lead/variables.tf
```diff
@@ -18,6 +18,8 @@ variable "enable_keycloak" {}
 
 variable "enable_harbor" {}
 
+variable "enable_artifactory_jcr" {}
```
mrparkers, line 21: let's add `default = false` here

stages/config/lead/variables.tf
```diff
@@ -26,6 +28,10 @@ variable "harbor_hostname" {
   description = "Harbor hostname provided by app stage"
 }
 
+variable "artifactory_jcr_hostname" {
+  description = "Artifactory hostname provided by app stage"
```
mrparkers, line 32: let's add a default here too so we don't have to provide this if artifactory isn't enabled

stages/apps/lead/variables.tf
mrparkers: still looking to get this resolved

mrparkers (approved):

PR #447, opened 2021-03-24
customize name of mattermost sparky

mrparkers:


RobertKelly (approved):
Helps make it generic!

PR #448, opened 2021-03-24
Add Velero backup 

brentwolfram:
Added S3 bucket, user, user policy, access key and Helm release for Velero. Created daily backup schedule for Harbor with 72hrs retention. Backup schedules (separate Velero API type) are contained within a separate chart (velero-schedule).


stages/cloud-provider/aws/lead/variables.tf
```diff
@@ -101,4 +101,11 @@ locals {
 
 variable "docker_registry_mirror" {
   default = ""
+}
+
+variable "velero_user" {
+  default = "velero"
+}
+
+variable "enable_velero" {
```
mrparkers, line 110: I think we should provide a default for this.  probably `false`

modules/tools/velero/main.tf
```diff
@@ -0,0 +1,37 @@
...
+  repository = "vmware-tanzu"
+  name       = "velero"
+  chart      = "vmware-tanzu/velero"
+  namespace  = module.velero_namespace.name
+  version    = "2.15.0"
+
+  values = [
+    templatefile("${path.module}/velero-values.tpl", {
+      velero_accesskey_id     = var.velero_aws_access_key_id
+      velero_accesskey_secret = var.velero_aws_secret_access_key
```
mrparkers, line 20: secrets should be set using `set_sensitive`, otherwise they will appear in plaintext in the console running terraform.  https://registry.terraform.io/providers/hashicorp/helm/latest/docs/resources/release#set_sensitive

modules/tools/velero/charts/velero-schedule/templates/harbor-daily-backup.yaml
```diff
@@ -0,0 +1,42 @@
+apiVersion: velero.io/v1
+kind: Schedule
+metadata:
+  name: harbor-daily-backup
```
mrparkers, line 4: does it make sense to include this within the the `velero` module, or should it be included within the `harbor` module if `enable_velero` is set to `true`?  I think this chart shouldn't be installed if `enable_velero` is `true` and `enable_harbor` is false.

mrparkers (changes requested):

modules/tools/velero/velero-values.tpl
```diff
@@ -0,0 +1,35 @@
...
+    cloud: |
+      [default]
+      aws_access_key_id=${velero_accesskey_id}
+      aws_secret_access_key=${velero_accesskey_secret}
+
+serviceAccount:
+  server:
+    create: true
+    name: velero
+    annotations:
```
mrparkers, line 35: we don't have to do this now, but we should be using EKS IAM roles for service accounts (https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) to provide AWS creds to workloads running within EKS (assuming velero supports it).  we currently do this with cert-manager and external-dns, although we don't do it for Vault because they didn't support IRSA at the time.

stages/cloud-provider/aws/lead/variables.tf
brentwolfram: Agreed, will update.

modules/tools/velero/main.tf
brentwolfram: Have switched to multi-line string using heredoc syntax using set_sensitive, as we discussed.

modules/tools/velero/velero-values.tpl
brentwolfram: I got this working but only after adding "fsGroup: 65534" to the security context in values for the Velero helm release. There is an issue (https://github.com/vmware-tanzu/velero/issues/2981) that apparently is resolved in K8s 1.19 (https://github.com/kubernetes/kubernetes/issues/82573). I’ve removed the user, access key, etc..that I previously created.

modules/tools/velero/charts/velero-schedule/templates/harbor-daily-backup.yaml
brentwolfram: Added a list to control what namespaces Velero is used on. I added a feature flag in the Velero module that uses contains on this list to decide whether the Helm release for the chart containing the backup schedules for that namespace is installed. The Helm release per namespace contains one or more instances of “Schedule” API type, specific to that namespace. So we can still have different label targets, pre-/post-hooks, etc…for workloads in the namespace. Not perfect, but OK for now since we are only interested in two namespaces. Open to other ideas.

ChrisSchreiber (approved):

mrparkers (approved):

PR #452, opened 2021-03-29
add image pull secrets to product-jenkins

mrparkers:
turns out we can just use the docker config json we were already building for this


modules/lead/product-jenkins/dockercfg.tf
```diff
@@ -34,3 +34,31 @@ resource "kubernetes_secret" "jenkins_repository_dockercfg" {
...
+
+  data = {
+    ".dockerconfigjson" = data.template_file.dockercfg.rendered
+  }
+}
+
+resource "kubernetes_secret" "staging_image_pull_secrets" {
+  provider = kubernetes.staging
+  metadata {
+    name      = "image-pull-secret"
```
gesparza3, line 55: So to get his working we'll just have to have our deployment point to this new secret?

mrparkers: that's the idea

gesparza3 (approved):
resources looks good to me, approving now. Should probably wait to merge until functionality is tested

RobertKelly (approved):

modules/lead/product-jenkins/pipelines-github.tpl
```diff
@@ -0,0 +1,22 @@
+{"jobs":[%{ for name,pipeline in pipelines ~}{"script":"
+    folder('${pipeline.org}') {
+    }
+    multibranchPipelineJob('${pipeline.org}/${pipeline.repo}') {
+        triggers {
+            periodic(1)
+        }
+        branchSources {
+            github {
```
RobertKelly, line 9: Looks good

modules/lead/product-jenkins/variables.tf
```diff
@@ -34,3 +34,6 @@ variable "pipelines" {
 
 variable "vault_namespace" {}
 variable "vault_root_token_secret" {}
+variable "jenkins_pipeline_source" {
+  default = "git"
```
RobertKelly, line 38: Saw `github` used earlier. Is `git` what we want to default to?

mrparkers: yeah, that's what we will continue to use for ignite labs

PR #453, opened 2021-03-30
Eng 2049 - Manual SDM product terraform state

PaulDHenson:
- Added logic to the vault provider so that it will continue to use the internal cluster url by default, but if a url is being provided it will use that instead.
- Added a variable called vault_path_url to the product-jenkins and product-base modules as well as the product-jenkins stack so that we can pass that information all the way to the product-base module.
- Created a config_path variable and defined the default location


modules/lead/product-base/variables.tf
```diff
@@ -3,3 +3,6 @@ variable "product_name" {}
 variable "image_whitelist" {}
 variable "vault_namespace" {}
 variable "vault_root_token_secret" {}
+variable "vault_path_url" {
```
mrparkers, line 6: I think `vault_address` would be a better name for this, since that's how it's used

modules/lead/product-base/variables.tf
```diff
@@ -3,3 +3,6 @@ variable "product_name" {}
...
+    default = ""
```
mrparkers, line 7: why not set the default to `http://vault.${var.vault_namespace}.svc.cluster.local` since that's effectively how you're using it?

stacks/product-jenkins/variables.tf
```diff
@@ -16,6 +16,10 @@ variable "config_context" {
   default = null
 }
 
+variable "config_path" {
+  default = "~/.kube/config"
```
mrparkers, line 20: I haven't tested this, but I have a feeling that this is going to break our usage of the jenkins product from within the cluster, since the product operator isn't using a kubeconfig to authenticate.  we should be really careful with the defaults we set here.

mrparkers (changes requested):

modules/lead/product-jenkins/variables.tf
```diff
@@ -34,3 +34,6 @@ variable "pipelines" {
 
 variable "vault_namespace" {}
 variable "vault_root_token_secret" {}
+variable "vault_adress" {
+  default = "http://vault.${var.vault_namespace}.svc.cluster.local"
```
ChrisSchreiber, line 38: you cannot use a variable as part of the default value for another variable

modules/lead/product-base/variables.tf
PaulDHenson: This is a good point and I just pushed those changes so I'll mark this as resolved.

modules/lead/product-base/variables.tf
PaulDHenson: This is a much better name, so I've renamed it and I'll mark this as resolved.

modules/lead/product-jenkins/variables.tf
mrparkers: my bad, thought you could

ChrisSchreiber (approved):

mrparkers (approved):

PR #461, opened 2021-04-29
Add internal ingress

jknight-liatrio:
This creates an nginx-ingress-controller to handle ingresses that are specifically accessible by the internal VPN.

### DIFF
--- a/modules/common/nginx-ingress/main.tf
+++ b/modules/common/nginx-ingress/main.tf
@@ -1,17 +1,3 @@
-data "template_file" "nginx_ingress_values" {
-  count    = var.enabled ? 1 : 0
-  template = file("${path.module}/nginx-ingress-values.tpl")
-
-  vars = {
-    ingress_controller_type         = var.ingress_controller_type
-    ingress_class                   = var.ingress_class
-    ingress_external_traffic_policy = var.ingress_external_traffic_policy
-    service_account                 = var.service_account
-    cluster_wide                    = var.cluster_wide
-    default_certificate             = var.default_certificate
-  }
-}
-
 resource "helm_release" "nginx_ingress" {
   count      = var.enabled ? 1 : 0
   repository = "https://charts.helm.sh/stable"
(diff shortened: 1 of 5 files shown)


ChrisSchreiber (approved):
LGTM

modules/common/nginx-ingress/main.tf
```diff
@@ -21,6 +7,17 @@ resource "helm_release" "nginx_ingress" {
...
   timeout    = 600
 
-  values = [data.template_file.nginx_ingress_values[0].rendered]
+  values = [
+    templatefile("${path.module}/nginx-ingress-values.tpl", {
+      ingress_controller_type             = var.ingress_controller_type
+      ingress_class                       = var.ingress_class
+      ingress_external_traffic_policy     = var.ingress_external_traffic_policy
+      service_account                     = var.service_account
+      service_annotations                 = var.service_annotaitons
```
mrparkers, line 16: ~`annotaitons`~ -> `annotations`

modules/common/nginx-ingress/nginx-ingress-values.tpl
```diff
@@ -24,6 +24,10 @@ controller:
   service:
     type: ${ingress_controller_type}
     externalTrafficPolicy: ${ingress_external_traffic_policy}
+    annotations:
+      ${indent( 6, yamlencode( service_annotations ) ) }
```
mrparkers, line 28: ~`annotaitons`~ -> `annotations`

modules/lead/toolchain-ingress/main.tf
```diff
@@ -165,3 +165,58 @@ module "toolchain_ingress" {
...
+  source                          = "../../common/nginx-ingress"
+  namespace                       = var.namespace
+  name                            = "internal"
+  ingress_controller_type         = var.ingress_controller_type
+  ingress_external_traffic_policy = var.ingress_external_traffic_policy
+  ingress_class                   = "internal-nginx"
+  service_annotaitons = {
+    "service.beta.kubernetes.io/aws-load-balancer-internal": true
+  }
+  service_load_balancer_source_ranges = ["10.1.32.0/20"]
```
mrparkers, line 218: the client cidr block is actually `10.1.48.0/20`: https://github.com/liatrio/aws-terraform/blob/master/environments/sharedsvc/network/internal-vpn.tf#L63

ChrisSchreiber: I tested both... `10.1.32.0/20` works and `10.1.48.0/20` doesn't. I think we ran into this before and I think `10.1.48.0/20` is the address in shared services but the traffic is coming from a different ip range because of the cross account vpc configuration stuff. It is also documented as `VPC for internal VPN (split tunnel vpn peered to other subnets)` in the aws-terraform readme so i'm pretty sure it's the right one.

PR #472, opened 2021-06-21
fix vault connection in product-base to use port

mrparkers:


ChrisSchreiber (commented):
Should the port also be added to external address "https://vault.${var.vault_namespace}.${var.cluster_domain}"?

mrparkers:
That shouldn't be needed, since the external connection goes through nginx over the standard ports you'd expect.  Port 8200 is only needed when talking to vault directly (within the cluster)

ChrisSchreiber (approved):

PR #476, opened 2021-07-22
Sonarqube ingress

ChrisSchreiber:


modules/tools/sonarqube/variables.tf
```diff
@@ -1,6 +1,29 @@
-variable "enable_sonarqube" {
-  default = true
+variable "admin_password" {
+  type      = string
+  sensitive = true
+}
+
+variable "force_authentication" {
+  default     = true
+  description = "Require authentication. Set to false to enable anoymouse access."
```
mrparkers, line 8:
```suggestion
  description = "Require authentication. Set to false to enable anonymous access."
```

mrparkers: also, this variable doesn't appear to be used.  the chart values hardcode `sonar.forceAuthentication` to `true`

PR #478, opened 2021-08-06
feat: upgrade vault, automate initialization process

mrparkers:
quite a few things happening in this PR:

- bump Vault chart to the latest version, installing the latest version of Vault
- update authentication method for Vault to AWS to switch to IRSA instead of passing credentials between stages
- remove our usage of the shell resource module for the initialization process, which was very error prone and constantly trying to update with each run of `terragrunt apply`.
- switch to the (forked) [vault-init](https://github.com/liatrio/vault-init) project to help with automation of vault's initialization process


modules/tools/vault-less-secure/values.tpl
```diff
@@ -33,6 +31,22 @@ server:
     limits:
       memory: 256Mi
       cpu: 512m
+  serviceAccount:
+    create: false
+    name: vault
+
+  extraContainers:
+    - name: vault-init
+      image: harbor.parker.gg/library/vault-init:v1
```
mrparkers, line 40: todo: replace with actual docker image once https://github.com/liatrio/vault-init/pull/1 is merged and released

stages/apps/lead/sonarqube.tf
```diff
@@ -1,13 +1,14 @@
 data "vault_generic_secret" "sonarqube" {
-  path = "lead/aws/${data.aws_caller_identity.current.account_id}/sonarqube"
+  count = var.enable_sonarqube ? 1 : 0
+  path  = "lead/aws/${data.aws_caller_identity.current.account_id}/sonarqube"
```
mrparkers, line 3: fix: don't query vault for this secret when `enable_sonarqube` is set to `false`

PR #479, opened 2021-08-09
Adding sonarqube to shared services

steeeveo:
* Adds sonarqube deployment to shared services cluster
* Handles some errors from the latest AMI version used by the cluster
* Adds an additional external-dns release to handle the public zone
* Adds external-facing ingress components
* Upgrades the external-dns release to the latest version

### DIFF
--- a/modules/environment/aws/eks/main.tf
+++ b/modules/environment/aws/eks/main.tf
@@ -9,7 +9,7 @@ systemctl enable amazon-ssm-agent
 
 echo '{"registry-mirrors": [${ var.docker_registry_mirror != "" ? format("\"%s\"", var.docker_registry_mirror) : ""}]}' | cat /etc/docker/daemon.json - | jq -s '.[0] * .[1]' > /tmp/daemon.json
 mv /tmp/daemon.json /etc/docker/daemon.json
-systemctl reload docker
+systemctl restart docker
 EOF
 
   tags = {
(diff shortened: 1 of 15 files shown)


stages/apps/sharedsvc/sonarqube.tf
```diff
@@ -0,0 +1,22 @@
+data "vault_generic_secret" "sonarqube" {
+  path = "lead/aws/265560927720/sonarqube"
```
mrparkers, line 2: we should derive the account number from the [`caller_identity` data source](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/caller_identity)

modules/environment/aws/eks/main.tf
```diff
@@ -9,7 +9,7 @@ systemctl enable amazon-ssm-agent
 
 echo '{"registry-mirrors": [${ var.docker_registry_mirror != "" ? format("\"%s\"", var.docker_registry_mirror) : ""}]}' | cat /etc/docker/daemon.json - | jq -s '.[0] * .[1]' > /tmp/daemon.json
 mv /tmp/daemon.json /etc/docker/daemon.json
-systemctl reload docker
+systemctl restart docker
```
mrparkers, line 12: let's revert this change, we want this to be changed in the upcoming PR to upgrade the LEAD clusters to 1.17

mrparkers (changes requested):

stages/apps/sharedsvc/sonarqube.tf
steeeveo: The hardcoding was intentional; as we didn't see a reason to make it dynamic, but I'm fine making it dynamic.

mrparkers (approved):

PR #480, opened 2021-08-10
update EKS module version and fix issue with docker service in bootstrap script

mrparkers:


stages/cloud-provider/aws/lead/outputs.tf
```diff
@@ -93,6 +93,7 @@ output "vault_aws_access_key_id" {
 
 output "vault_aws_secret_access_key" {
   value = aws_iam_access_key.vault_key.secret
+  sensitive = true
```
steeeveo, line 96: Will this be applied with terraform 0.15+?

mrparkers: Yes, at this point I'm running Terraform v1.0.1, so everyone else will be forced to upgrade to at least that version anyways.

steeeveo (approved):

PR #484, opened 2021-08-12
Github runners in shared services

steeeveo:
* copies github runner configuration from lead to shared services

### DIFF
--- a/modules/environment/aws/github-runners-s3/main.tf
+++ b/modules/environment/aws/github-runners-s3/main.tf
@@ -1,8 +1,12 @@
 data "aws_caller_identity" "current" {
 }
 
+locals {
+  bucket_name = "github-runners-${data.aws_caller_identity.current.account_id}-${var.cluster_name}.liatr.io"
+}
+
 resource "aws_s3_bucket" "github-runner" {
-  bucket = "github-runners-${data.aws_caller_identity.current.account_id}-${var.cluster_name}.liatr.io"
+  bucket = local.bucket_name
   tags = {
     Name      = "Github Runner States"
     ManagedBy = "Terraform https://github.com/liatrio/lead-terraform"
@@ -49,7 +53,7 @@ resource "aws_iam_policy" "github_runners" {
                 "s3:GetBucketVersioning",
                 "s3:CreateBucket"
      ],
-     "Resource": ["arn:aws:s3:::github-runners-${data.aws_caller_identity.current.account_id}-${var.cluster_name}.liatr.io"]
+     "Resource": ["arn:aws:s3:::${local.bucket_name}"]
    },
    {
      "Effect": "Allow",
@@ -58,7 +62,7 @@ resource "aws_iam_policy" "github_runners" {
                 "s3:GetObject",
                 "s3:DeleteObject"
      ],
-     "Resource": ["arn:aws:s3:::github-runners-${data.aws_caller_identity.current.account_id}-${var.cluster_name}.liatr.io/*"]
+     "Resource": ["arn:aws:s3:::${local.bucket_name}/*"]
    },
    {
      "Effect": "Allow",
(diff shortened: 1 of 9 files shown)


stages/apps/sharedsvc/variables.tf
```diff
@@ -54,4 +54,31 @@ variable "vault_address" {
   default = "https://vault.internal.services.liatr.io"
 }
 
+variable "github_runners_service_account_arn" {
+}
+
 variable "iam_caller_identity_headers" {}
+
+# example_value = {
```
jknight-liatrio, line 62: Is there a reason we left these comments in?

stages/cloud-provider/aws/sharedsvc/variables.tf
```diff
@@ -75,3 +75,9 @@ variable "cluster_domain" {
 variable "internal_cluster_domain" {
   default = "internal.services.liatr.io"
 }
+
+variable "github_runner_service_accounts" {
+  description = "a list of service accounts to apply IRSA trust policy to. each service account should be in the format `namespace:name`"
```
jknight-liatrio, line 80: If these require some special formatting, why not use a list of objects? Just curious, it's probably easier just to pass them through if this is the way that the github runner module consumes them.

jknight-liatrio (approved):

stages/cloud-provider/aws/sharedsvc/variables.tf
steeeveo: I'm not entirely sure. I think `namespace:name` is the format that the iam role expects; I suppose it was easier to specify a string to include than to pass a map that gets formatted in the final resource.

stages/apps/sharedsvc/variables.tf
steeeveo: The object structure is somewhat complex and not immediately obvious; it would be nice if we could make it so the comments aren't needed, I agree.

steeeveo: note: tried converting the complex map to a set of objects, but terraform's `for_each` only supports maps or sets of strings

modules/tools/vault-less-secure/main.tf
```diff
@@ -61,7 +61,7 @@ resource "helm_release" "vault" {
       vault_hostname                = var.vault_hostname
       vault_service_account_arn     = var.vault_service_account_arn
       vault_credentials_secret_name = local.vault_secret_name
-      vault_config                  = indent(6, templatefile("${path.module}/vault-config.hcl.tpl", {
+      vault_config = indent(6, templatefile("${path.module}/vault-config.hcl.tpl", {
```
mrparkers, line 64: let's run `terraform fmt` to undo some of these whitespace changes

steeeveo: this was the result of `terraform fmt`

modules/environment/aws/github-runners-s3/main.tf
```diff
@@ -49,7 +53,7 @@ resource "aws_iam_policy" "github_runners" {
                 "s3:GetBucketVersioning",
                 "s3:CreateBucket"
      ],
-     "Resource": ["arn:aws:s3:::github-runners-${data.aws_caller_identity.current.account_id}-${var.cluster_name}.liatr.io"]
+     "Resource": ["arn:aws:s3:::${local.bucket_name}"]
```
mrparkers, line 56: this arn string could be replaced with `aws_s3_bucket.github-runner.arn` so we don't have to build the string ourselves.

modules/environment/aws/github-runners-s3/main.tf
```diff
@@ -58,7 +62,7 @@ resource "aws_iam_policy" "github_runners" {
                 "s3:GetObject",
                 "s3:DeleteObject"
      ],
-     "Resource": ["arn:aws:s3:::github-runners-${data.aws_caller_identity.current.account_id}-${var.cluster_name}.liatr.io/*"]
+     "Resource": ["arn:aws:s3:::${local.bucket_name}/*"]
```
mrparkers, line 65: same as above

mrparkers (approved):

PR #487, opened 2021-08-19
Adding Rode to shared services

steeeveo:
* Updates rode module to use latest helm chart
* Calls rode module from sharedsvc
* Configures OIDC client in sharedsvc keycloak


stages/cloud-provider/aws/sharedsvc/iam.tf
```diff
@@ -54,6 +54,63 @@ module "cert_manager_iam" {
   openid_connect_provider_url = module.eks.aws_iam_openid_connect_provider.url
 }
 
+resource "aws_iam_role" "rode_service_account" {
```
steeeveo, line 57: `cloud-provider` has already been applied to sharedsvc, but I wonder if this should be removed? I'm not seeing where it would be used.

modules/tools/rode/main.tf
```diff
@@ -1,19 +1,73 @@
...
+    templatefile("${path.module}/rode-ui-values.tpl", {
+      ingress_enabled  = true
+      ingress_hostname = var.ui_ingress_hostname
+      ingress_annotations = {
+        "nginx.ingress.kubernetes.io/proxy-buffer-size" : "8k"
+        "kubernetes.io/ingress.class" : var.ingress_class
+      }
+
+      oidc_config = {
+        enabled : var.oidc_issuer_url == "" ? false : true,
```
alexashley, line 67:
```suggestion
        enabled : var.oidc_issuer_url != "",
```

modules/tools/rode/main.tf
```diff
@@ -1,19 +1,73 @@
...
+        roleClaimPath : "resource_access.${var.oidc_issuer_client_id}.roles",
+        tlsInsecureSkipVerify : false
+      }
+
+      grafeas_image_tag = var.grafeas_image_tag
     })
   ]
 }
 
+resource "helm_release" "rode-ui" {
```
alexashley, line 41:
```suggestion
resource "helm_release" "rode_ui" {
```

alexashley: Also I think we'll want a `depends_on` for this resource

```hcl
depends_on = [
    helm_release.rode,
]
```

modules/tools/rode/rode-values.tpl
```diff
@@ -1,30 +1,21 @@
...
-rbac:
-  serviceAccountAnnotations:
-    eks.amazonaws.com/role-arn: ${iam_arn}
+rode-ui:
+  enabled: false
 
-localstack:
-  enabled: ${localstack_enabled}
 
-resources:
```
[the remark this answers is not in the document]
steeeveo: `rode-ui` doesn't specify defaults, but the others do. Do you think it would be better to set that as a default in the chart, or set it here? I'd lean towards setting it in the chart; I wouldn't expect everybody consuming the chart to figure out limits on their own?

modules/tools/rode/main.tf
```diff
@@ -1,19 +1,71 @@
...
-      ingress_hostname   = "rode.${var.namespace}.${var.cluster}.${var.root_zone_name}"
-      localstack_enabled = var.localstack_enabled
+      ingress_enabled  = true
+      ingress_hostname = var.rode_ingress_hostname
+      ingress_annotations = {
+        "kubernetes.io/ingress.class" : var.ingress_class
+      }
+
+      oidc_config = {
+        enabled : var.oidc_issuer_url == "" ? false : true,
```
alexashley, line 28:
```suggestion
        enabled : var.oidc_issuer_url != "",
```

modules/tools/rode/rode-values.tpl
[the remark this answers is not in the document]
alexashley: Interesting, I didn't realize we had defaults in the chart. Usually charts leave resources unset, but if we're already specifying them then I agree it makes sense to set them for rode-ui in the chart as well.

modules/tools/rode/variables.tf
```diff
@@ -1,15 +1,36 @@
...
 
-variable "namespace" {}
+variable "oidc_issuer_url" {
+  description = "Leave blank to disable OIDC"
+  default     = ""
+}
+variable "oidc_issuer_client_id" {
+  default = ""
+}
+variable "oidc_issuer_client_secret" {
```
alexashley, line 21:
```suggestion
variable "oidc_client_secret" {
```

modules/tools/rode/variables.tf
```diff
@@ -1,15 +1,36 @@
...
+variable "ingress_class" {
+  default = "toolchain-nginx"
+}
 
-variable "namespace" {}
+variable "oidc_issuer_url" {
+  description = "Leave blank to disable OIDC"
+  default     = ""
+}
+variable "oidc_issuer_client_id" {
```
alexashley, line 18: Nitpick, but I'd just call this client id/client secret. 

```suggestion
variable "oidc_client_id" {
```

alexashley (commented):
Just one nitpick, otherwise LGTM

modules/tools/rode/main.tf
```diff
@@ -1,19 +1,71 @@
...
+        tlsInsecureSkipVerify : false
+      }
+
+      grafeas_image_tag = var.grafeas_image_tag
     })
   ]
 }
 
+resource "helm_release" "rode_ui" {
+  count = var.ui_ingress_hostname == "" ? 0 : 1
```
ChrisSchreiber, line 41: Is there a use case for installing the UI without an ingress? I'm thinking if I want to test this module locally I might install it without an ingress and port forward the service.

steeeveo: Since the module is fairly purpose built I wouldn't think you'd want to use it outside of this architecture, and I thought we weren't really trying to make this config local-friendly, but I'm not opposed to switching back to an `enabled` variable for that flag. Thoughts?

ChrisSchreiber: Yeah we probably will not use this module outside of the shared svc stage env but if we are going to have a feature flag here I think it should be purpose specific and enabling/disabled just the ui ingress on the hostname separately.

ChrisSchreiber (approved):

PR #489, opened 2021-08-23
Include tfsec collector in Rode install

alexashley:
~I haven't applied this yet, but wanted to open this before I did.~ I've already applied these changes. 
 Also includes a few other changes:

- excludes Rode namespace from downscaling
- force SSL redirects on all ingress
- fixes the Rode UI count

### DIFF
--- a/modules/tools/rode/main.tf
+++ b/modules/tools/rode/main.tf
@@ -25,7 +25,8 @@ resource "helm_release" "rode" {
       ingress_enabled     = true
       ingress_hostname    = var.rode_ingress_hostname
       ingress_annotations = {
-        "kubernetes.io/ingress.class" : var.ingress_class
+        "kubernetes.io/ingress.class" : var.ingress_class,
+        "nginx.ingress.kubernetes.io/force-ssl-redirect": "true",
       }
 
       oidc_config = {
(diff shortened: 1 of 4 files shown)


modules/tools/rode/main.tf
```diff
@@ -38,7 +43,7 @@ resource "helm_release" "rode" {
 }
 
 resource "helm_release" "rode_ui" {
-  count = var.rode_ui_enabled
+  count = var.rode_ui_enabled ? 1 : 0
```
steeeveo, line 46: thanks for catching this 😅

modules/tools/rode/tfsec-collector-values.yaml.tpl
```diff
@@ -0,0 +1,19 @@
+rode:
+  host: rode.${namespace}.svc.cluster.local:50051
```
steeeveo, line 2: They should be in the same namespace, right? idk which is better practice, but I know the UI just uses `rode` to get to the rode service.

modules/tools/rode/variables.tf
```diff
@@ -38,3 +38,8 @@ variable "grafeas_elasticsearch_password" {
 variable "grafeas_image_tag" {
   default = "v0.8.8"
 }
+
+variable "tfsec_collector_hostname" {
```
steeeveo, line 42: should this be set in `stages/apps/sharedsvc/rode.tf` or does that come later?

modules/tools/rode/tfsec-collector-values.yaml.tpl
alexashley: Probably doesn't make a different, but I thought using the FQDN saved some extra DNS lookups. Couldn't find a source, but tried on a pod:

```shell
/usr/src/app # nslookup rode.rode-demo.svc.cluster.local
Server:		10.43.0.10
Address:	10.43.0.10:53

Name:	rode.rode-demo.svc.cluster.local
Address: 10.43.92.233

/usr/src/app # nslookup rode
Server:		10.43.0.10
Address:	10.43.0.10:53

Name:	rode.rode-demo.svc.cluster.local
Address: 10.43.92.233

** server can't find rode.svc.cluster.local: NXDOMAIN

** server can't find rode.cluster.local: NXDOMAIN

** server can't find rode.svc.cluster.local: NXDOMAIN

** server can't find rode.cluster.local: NXDOMAIN
```

steeeveo (approved):

Pactionly (approved):
LGTM

PR #490, opened 2021-08-24
Add Rode terraform client and configure gRPC ingress

alexashley:


steeeveo (approved):

modules/tools/rode/main.tf
```diff
@@ -22,11 +22,23 @@ resource "helm_release" "rode" {
 
   values = [
     templatefile("${path.module}/rode-values.tpl", {
-      ingress_enabled     = true
-      ingress_hostname    = var.rode_ingress_hostname
-      ingress_annotations = {
-        "kubernetes.io/ingress.class" : var.ingress_class,
-        "nginx.ingress.kubernetes.io/force-ssl-redirect": "true",
+      ingress = {
+        enabled = true
```
Pactionly, line 26: Not much context here, feel free to ignore me, but is there any value to having these enable-able separately?

alexashley: Potentially, although at the moment we've got callers that use both gRPC (the TF provider) and HTTP (pipelines using `curl` for `EvaluatePolicy`). So at least for the foreseeable future, we'll likely have both enabled. But it would be straightforward enough to add a check here, or just drive the enabled off of whether the host is set.

Pactionly (approved):

RobertKelly (approved):

PR #497, opened 2021-08-31
Adding harbor to shared svc

djtjwillia:


stages/apps/sharedsvc/outputs.tf
```diff
@@ -18,6 +18,14 @@ output "keycloak_hostname" {
   value = module.keycloak.keycloak_hostname
 }
 
+output "harbor_hostname" {
+  value = module.harbor.hostname
```
mrparkers, line 22: `module.harbor` has the `count` argument now, so I think we'll have to do this instead:

```suggestion
  value = module.harbor[0].hostname
```

stages/apps/lead/harbor.tf
```diff
@@ -1,15 +1,24 @@
...
 data "vault_generic_secret" "harbor" {
   path = "lead/aws/${data.aws_caller_identity.current.account_id}/harbor"
 }
 
 module "harbor" {
   source = "../../../modules/tools/harbor"
 
-  enable                       = var.enable_harbor
-  cluster                      = var.cluster_name
+  count                        = var.enable_harbor ? 1 : 0
```
mrparkers, line 16: this is a good change, but any outputs for this stage that used to reference `module.harbor` now need to reference `module.harbor[0]`.

stages/apps/lead/harbor.tf
```diff
@@ -1,15 +1,24 @@
+locals {
+  ingress_annotations = {
```
mrparkers, line 2: I wonder if these should be moved somewhere more generic, since they could be used for other deployments as well.  If the intention is to only use these annotations for the harbor ingress, then this should probably be renamed to `harbor_ingress_annotations`, since this local variable can be accessed within other files in this module as well.

stages/apps/sharedsvc/harbor.tf
```diff
@@ -0,0 +1,35 @@
+locals {
+  ingress_annotations = {
+    "nginx.ingress.kubernetes.io/force-ssl-redirect" : true
+    "nginx.ingress.kubernetes.io/proxy-body-size" : "0"
+    "kubernetes.io/ingress.class" : module.nginx_external.ingress_class
+  }
```
mrparkers, line 6: same as above, I think these should be moved to a more generic place or renamed so it's understood that these are specifically for harbor

stages/apps/lead/harbor.tf
djtjwillia: That is a really good point. Yeah we will likely have ingress annotations for various apps, but they will usually be specific, so ill need to update this.

stages/apps/lead/harbor.tf
djtjwillia: good catch, resolving this in the outputs

steeeveo (approved):

PR #499, opened 2021-09-02
ENG-2075: Deploy Rode SonarQube Collector

alexashley:


modules/tools/rode/main.tf
```diff
@@ -111,6 +111,33 @@ resource "helm_release" "rode_tfsec_collector" {
...
+  wait       = true
+
+  set_sensitive {
+    name  = "rode.auth.oidc.clientSecret"
+    value = var.collector_client_secret
+  }
+
+  values = [
+    templatefile("${path.module}/sonar-collector-values.yaml.tpl", {
+      oidc_auth_enabled = var.oidc_token_url != ""
```
steeeveo, line 129: At one point we talked about making all of these 'enabled' flags explicit instead of implied from the existence of the token url; it looks like we made it explicit for sonarqube but implicit for rode. I could go either way, but wanted to bring it up so we could standardize one way or the other.

steeeveo (approved):

modules/tools/rode/main.tf
alexashley: For now, I just updated it to match the others that use that local.

PR #500, opened 2021-09-02
Deploy Harbor collector

alexashley:
It's worth mentioning that this isn't the latest and greatest version of the collector, instead it's a patch version behind so that it's compatible with the version of Harbor that's in use. See https://github.com/rode/collector-harbor/pull/19 for more details. 

If we upgrade Harbor, we'll need to bump the collector as well.

### DIFF
--- a/stages/apps/sharedsvc/outputs.tf
+++ b/stages/apps/sharedsvc/outputs.tf
@@ -38,6 +38,10 @@ output "sonarqube_collector_url" {
   value = "http://rode-collector-sonarqube.${module.rode_namespace.name}.svc.cluster.local/webhook/event"
 }
 
+output "harbor_collector_url" {
+  value = "http://rode-collector-harbor.${module.rode_namespace.name}.svc.cluster.local/webhook/event"
+}
+
 output "rode_oidc_client_id" {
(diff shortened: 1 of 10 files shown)


stages/apps/sharedsvc/outputs.tf
```diff
@@ -38,6 +38,10 @@ output "sonarqube_collector_url" {
   value = "http://rode-collector-sonarqube.${module.rode_namespace.name}.svc.cluster.local/webhook/event"
 }
 
+output "harbor_collector_url" {
+  value = "http://rode-collector-harbor.${module.rode_namespace.name}.svc.cluster.local/webhook/event"
```
RobertKelly, line 42: Why are these not https?  Just curious.

alexashley: It's all traffic inside the Kubernetes cluster, neither of these collectors are exposed externally. 

If we did want that, we could update the collectors to serve HTTPS or use something like Istio to get mTLS

RobertKelly: Something for later thanks!

RobertKelly (approved):

PR #504, opened 2021-09-09
fix: updating vars for prometheus in shared svc

djtjwillia:


stages/apps/sharedsvc/variables.tf
```diff
@@ -19,12 +22,19 @@ variable "enable_autoscaler_scale_down" {
   default = true
 }
 
+variable "enable_dashboard" {
+  default = true
```
steeeveo: I think we wanted to default all the enable flags to false, to make it easier to see what's enabled for a given environment in lead-environments

stages/apps/sharedsvc/kube-prometheus-stack.tf
```diff
@@ -6,16 +6,16 @@ module "monitoring_namespace" {
   source    = "../../../modules/common/namespace"
   namespace = var.monitoring_namespace
   annotations = {
-    name                                         = var.monitoring_namespace
-    cluster                                      = var.cluster_name
+    name    = var.monitoring_namespace
+    cluster = var.cluster_name
```
steeeveo: Is cluster_name needed for prometheus? Just wondering, as we'd managed to avoid adding it thus far

ChrisSchreiber (approved):

PR #506, opened 2021-09-10
update eks cluster to use managed node groups

mrparkers:


modules/affinity/essential-toleration-values/essential-toleration.tpl
```diff
@@ -3,7 +3,7 @@ affinity:
     requiredDuringSchedulingIgnoredDuringExecution:
       nodeSelectorTerms:
       - matchExpressions:
-        - key: "node.kubernetes.io/lifecycle"
+        - key: "node.liatr.io/lifecycle"
```
mrparkers, line 6: this label had to be updated because `kubernetes.io` is a reserved prefix

modules/environment/aws/eks/iam.tf
```diff
@@ -1,13 +1,3 @@
-resource "aws_iam_openid_connect_provider" "default" {
-  url = module.eks.cluster_oidc_issuer_url
-
-  client_id_list = [
-    "sts.amazonaws.com",
-  ]
-
-  thumbprint_list = ["9e99a48a9960b14926bb7f3b02e22da2b0ab7280"]
-}
-
```
mrparkers, line 10: removed this because we're setting `enable_irsa` to `true` now in the EKS module

modules/environment/aws/eks/main.tf
```diff
@@ -131,122 +135,98 @@ resource "aws_security_group" "elb" {
...
+    version                = var.cluster_version
+    disk_size              = var.root_volume_size
+    update_config          = {
+      max_unavailable_percentage = 50
+    }
 
-  workers_group_defaults = {
-    tags = [
-      {
-        "key"                 = "kubernetes.io/cluster-autoscaler/enabled"
```
mrparkers, line 164: managed node groups automatically add these tags / labels by default, no need for them anymore

modules/environment/aws/iam/cluster-autoscaler/main.tf
```diff
@@ -43,42 +43,13 @@ data "aws_iam_policy_document" "cluster_autoscaler" {
...
-
-  statement {
-    sid    = "clusterAutoscalerOwn"
-    effect = "Allow"
-
-    actions = [
       "autoscaling:SetDesiredCapacity",
       "autoscaling:TerminateInstanceInAutoScalingGroup",
-      "autoscaling:UpdateAutoScalingGroup",
+      "ec2:DescribeLaunchTemplateVersions"
```
mrparkers, line 48: cluster autoscaler IAM changes were updated as per https://docs.aws.amazon.com/eks/latest/userguide/cluster-autoscaler.html

modules/tools/cluster-autoscaler/cluster-autoscaler-values.tpl
```diff
@@ -13,7 +17,7 @@ autoDiscovery:
   clusterName: ${cluster}
   enabled: true
   tags:
-    - "kubernetes.io/cluster-autoscaler/enabled"
+    - "k8s.io/cluster-autoscaler/enabled"
```
mrparkers, line 20: one of the label changes that came with managed node groups

modules/tools/cluster-autoscaler/cluster-autoscaler-values.tpl
```diff
@@ -1,7 +1,11 @@
+image:
+  repository: k8s.gcr.io/autoscaling/cluster-autoscaler
+  tag: v1.21.0
```
mrparkers, line 3: the AWS docs for cluster autoscaler _claim_ that you're supposed to use a tag that reflects what version of k8s you're running (so in this case, `v1.17.0`), but I don't believe that to be true.  this configuration seems to work, but I want to test it out a bit more before we merge this

modules/environment/aws/eks/output.tf
```diff
@@ -2,8 +2,12 @@ output "cluster_id" {
...
 }
 
-output "aws_iam_openid_connect_provider" {
-  value = aws_iam_openid_connect_provider.default
+output "aws_iam_openid_connect_provider_arn" {
+  value = module.eks.oidc_provider_arn
+}
+
+output "aws_iam_openid_connect_provider_url" {
+  value = module.eks.cluster_oidc_issuer_url
```
mrparkers, line 10: all of the changes regarding these variables have to do with the EKS module creating the `aws_iam_openid_connect_provider` for us, and so the outputs are exposed a bit differently than they were before

stages/cloud-provider/aws/lead/variables.tf
```diff
@@ -30,7 +30,7 @@ variable "asg_max_size" {
 
 variable "instance_types" {
   type    = list(string)
-  default = ["m5.xlarge", "c5.xlarge", "m4.xlarge", "c4.xlarge", "t3.xlarge", "r5.xlarge"]
+  default = ["m5.xlarge", "m4.xlarge", "t3.xlarge", "r5.xlarge"]
```
mrparkers, line 33: for some reason, the new ASGs only consisted of `c5.xlarge` images.  AWS MNG docs recommend that each node group uses the same instance type (meaning a similar number of vCPUs and memory), so mixing these doesn't work very well.

we can re-add these later if we want different preemptible node groups to use different image types, but the `m` images have a good combination of CPU and memory so I just kept those in for now.

stages/cloud-provider/aws/lead/versions.tf
```diff
@@ -3,7 +3,7 @@ terraform {
   required_providers {
     aws = {
       source  = "hashicorp/aws"
-      version = "~> 3.43.0"
+      version = ">= 3.56.0"
```
mrparkers, line 6: required due to the EKS module update

modules/environment/aws/eks/main.tf
```diff
@@ -86,10 +93,7 @@ resource "aws_security_group" "worker" {
     to_port   = 22
     protocol  = "tcp"
 
-    cidr_blocks = [
-      data.aws_vpc.lead_vpc.cidr_block,
-      #"10.1.32.0/20", # internal VPN
-    ]
+    cidr_blocks = concat([data.aws_vpc.lead_vpc.cidr_block], var.enable_ssh_access ? [var.internal_vpn_subnet] : [])
```
ChrisSchreiber, line 96: i like this... nice thinking 👍

stacks/product-jenkins/versions.tf
```diff
@@ -13,12 +13,12 @@ terraform {
 
     harbor = {
       source  = "liatrio/harbor"
-      version = "= 0.3.3"
+      version = "0.4.0"
```
mrparkers, line 16: this actually fixes the delete jobs failing, so we can put off updating this provider until later

ChrisSchreiber (approved):

PR #513, opened 2021-09-22
Exclude prometheus from downscaler, setup internal ingress for alertmanager

MisterC500:
- Excluded Prometheus from downscaling 
- Add internal_ingress_hostname to local variables for alertmanager to be accessed internally. 
- Add title_link value in the template file for the kube-prometheus-stack 
- Increased Prometheus operator memory
- Add Prometheus rule for detecting oom kills

### DIFF
--- a/modules/tools/kube-downscaler/values.tpl
+++ b/modules/tools/kube-downscaler/values.tpl
@@ -5,7 +5,7 @@ image:
     %{~ if excluded_namespaces != "" ~}
     - --exclude-namespaces=${excluded_namespaces}
     %{~ endif ~}
-    - --exclude-deployments=kube-downscaler,metrics-server,cluster-autoscaler-aws-cluster-autoscaler,prometheus-kube-prometheus-stack-prometheus,kube-prometheus-stack-kube-state-metrics
+    - --exclude-deployments=kube-downscaler,metrics-server,cluster-autoscaler-aws-cluster-autoscaler
     - --default-uptime=${uptime}
     - --include-resources=deployments,statefulsets
 resources:
--- a/modules/tools/kube-downscaler/values.tpl
+++ b/modules/tools/kube-downscaler/values.tpl
@@ -5,7 +5,7 @@ image:
     %{~ if excluded_namespaces != "" ~}
     - --exclude-namespaces=${excluded_namespaces}
     %{~ endif ~}
-    - --exclude-deployments=kube-downscaler,metrics-server,cluster-autoscaler-aws-cluster-autoscaler
+    - --exclude-deployments=kube-downscaler,metrics-server,cluster-autoscaler-aws-cluster-autoscaler,prometheus-kube-prometheus-stack-prometheus,kube-prometheus-stack-kube-state-metrics
     - --default-uptime=${uptime}
     - --include-resources=deployments,statefulsets
 resources:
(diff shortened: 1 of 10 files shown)


modules/tools/kube-prometheus-stack/values.tpl
```diff
@@ -128,21 +135,21 @@ alertmanager:
...
+     #   receiver: slack
+     # - match:
+     #     namespace: lead-system
+     #   receiver: slack
+     # - match:
+     #     namespace: istio-system
+     #   receiver: slack
+     # - match:
+     #     namespace: elasticsearch
+     #   receiver: slack
```
mrparkers, line 152: if these aren't needed, can they be removed?

ChrisSchreiber, line 152: Let's just delete these if we are not using them

modules/tools/kube-downscaler/values.tpl
```diff
@@ -5,7 +5,7 @@ image:
     %{~ if excluded_namespaces != "" ~}
     - --exclude-namespaces=${excluded_namespaces}
     %{~ endif ~}
-    - --exclude-deployments=kube-downscaler,metrics-server,cluster-autoscaler-aws-cluster-autoscaler
+    - --exclude-deployments=kube-downscaler,metrics-server,cluster-autoscaler-aws-cluster-autoscaler,prometheus-kube-prometheus-stack-prometheus,kube-prometheus-stack-kube-state-metrics
```
ChrisSchreiber, line 8: I want the downscaler exceptions to be applied as annotations on the deployments instead of added to the downscaler configuration.

modules/tools/kube-prometheus-stack/main.tf
```diff
@@ -28,7 +28,9 @@ resource "helm_release" "kube_prometheus_stack" {
       prometheus_slack_webhook_url = var.prometheus_slack_webhook_url
       prometheus_slack_channel     = var.prometheus_slack_channel
       grafana_hostname             = var.grafana_hostname
-      ingress_annotations          = var.ingress_annotations
+      alertmanager_hostname        = var.alertmanager_hostname
+      grafana_ingress_annotations  = var.grafana_ingress_annotations
+      alertmanager_ingress_annotations = var.alertmanager_ingress_annotations
```
ChrisSchreiber, line 33: make sure to run this through `tf fmt`

ChrisSchreiber (changes requested):

stages/apps/lead/nginx.tf
```diff
@@ -8,3 +8,13 @@ module "app_nginx" {
   issuer_kind    = module.cluster_issuer.issuer_kind
   namespace      = module.toolchain_namespace.name
 }
+
+module "monitoring_nginx" {
```
mrparkers, line 12: what is this needed for?

MisterC500: We do not need that, forgot to take that out.

modules/tools/kube-prometheus-stack/values.tpl
```diff
@@ -67,6 +67,8 @@ prometheusOperator:
   configReloaderCpu: 100m
   configReloaderMemory: 25Mi
 prometheus:
+  annotations:
+    downscaler/exclude: "true"
```
mrparkers, line 71: do we want to add these annotations to alertmanager / grafana too?

MisterC500: We talked with Scribbles we decided we don't want grafana up after hours.

mrparkers: so then grafana, alertmanager, and prometheus-node-exporter would be scaled down after hours, is that the goal?

MisterC500: Yup, just having prometheus up.

mrparkers (approved):

ChrisSchreiber (approved):

PR #516, opened 2021-09-27
Slack Silence alert button

MisterC500:
- Add button to silence alerts in alertmanager from slack.


mrparkers:
can you give me an example of what this URL looks like?  the template is a bit difficult for me to understand

PR #528, opened 2021-10-15
ci: require PR titles to use conventional commit prefixes

alexashley:
Adds a GitHub Action workflow to enforce that pull request titles use [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/) prefixes. If a PR has a single commit, then it also requires the commit to follow the convention, since GitHub will suggest the commit for the title when squash merging. 

By default, it uses the standard list of prefixes, although that can be overwritten:

 - feat
 - fix
 - docs
 - style
 - refactor
 - perf
 - test
 - build
 - ci
 - chore
 - revert

The action also supports validating scopes, but I didn't configure that for now. If there's a convention we want to enforce, then it should be flexible enough to accommodate that. 

Here's a few pull requests that demonstrate the action:
- mix of failing/successful statuses when editing PR title: https://github.com/alexashley/lead-terraform/pull/1
- check that single commits fail without a prefix: https://github.com/alexashley/lead-terraform/pull/3

### DIFF
--- a/.github/workflows/conventional-commits.yml
+++ b/.github/workflows/conventional-commits.yml
@@ -16,3 +16,14 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
           validateSingleCommit: true
+          types: |
+            feat
+            fix
+            docs
+            style
+            refactor
+            perf
+            test
+            build
+            ci
+            revert


rcoy-v:
The "official" conventional commit spec lists `chore`, at least on the webpage, but the Angular spec/project that this originated from does not support `chore` and a lot of checking tools don't either. I would agree with why it was removed, since it quickly becomes a catch-all for any contribution. Does this particular tool introduced allow it?

alexashley:
@rcoy-v The list in the PR comment is what comes out of the box, sourced from this `commitizen`  (another one of these conventional commit tools) NPM package. Here's that [list](https://github.com/commitizen/conventional-commit-types/blob/master/index.json). 

The action has a [`types`](https://github.com/amannn/action-semantic-pull-request/blob/25e3275e19cd5b3577a99478084255065002897c/action.yml#L11), so we can limit which prefixes are allowed if that's desirable. I don't have a strong opinion on the use of `chore`.

PR #542, opened 2021-10-29
refactor: manage custom prometheus via helm chart

ChrisSchreiber:
replace the k8s manifests for create prometheus rules with a helm chart to avoid provider errors caused by CRD no existing.


modules/tools/kube-prometheus-stack/main.tf
```diff
@@ -34,3 +34,9 @@ resource "helm_release" "kube_prometheus_stack" {
     })
   ]
 }
+
+resource "helm_release" "prometheus_resources" {
+  chart = "${path.module}/chart"
+  name = "lead-prometheus-resources"
+  namespace = var.namespace
```
alexashley, line 41:
```suggestion
  namespace = var.namespace
  depends_on = [
      helm_release.kube_prometheus_stack,
  ]
```

modules/tools/kube-prometheus-stack/rules.tf
```diff
@@ -1,99 +0,0 @@
-resource "kubernetes_manifest" "prometheus_rule_kube_cluster-at-max-nodes" {
```
alexashley, line 1: There's another file in this module, `service-monitors.tf`, that's entirely commented out, but it also uses the `kubernetes_manifest` resource. That could probably be removed as well.

ChrisSchreiber: good catch @alexashley, I moved those into the helm chart also and added a feature flag since we want to add those after we update k8s + harbor

alexashley (approved):

PR #545, opened 2021-11-09
test: add cert-manager module test, add github actions test workflow

mrparkers:


.github/workflows/destroy-vcluster.yml
```diff
@@ -0,0 +1,40 @@
+name: "Destroy vcluster"
+
+on:
+  pull_request:
+    types:
+      - closed
+  push:
+    branches:
+      - master
```
ChrisSchreiber, line 9: will this make it run twice (on pr close and merge to master)?

.github/workflows/test.yml
```diff
@@ -0,0 +1,91 @@
...
+
+          cd lead-terraform
+          go test -v ./modules/./...
+
+      - name: Destroy vcluster
+        if: github.ref == 'refs/heads/master'
+        env:
+          VCLUSTER_NAME: ${{ steps.vcluster-name.outputs.name }}
+          TERRAGRUNT_IAM_ROLE: ${{ secrets.AWS_LEAD_TERRAFORM_TEST_ROLE }}
+        run: terragrunt destroy --terragrunt-working-dir ./lead-environments/aws/vcluster -auto-approve
```
ChrisSchreiber, line 91: Do we still need this with the other workflow?

mrparkers: Yes, this is so the vcluster that gets created for the run against `master` gets cleaned up.

.github/workflows/destroy-vcluster.yml
mrparkers: I messed this up, I meant to only run this on PR close

ChrisSchreiber (approved):

PR #563, opened 2021-12-08
feat: use EKS addons to manage installation of coredns

mrparkers:


modules/environment/aws/eks/addons.tf
```diff
@@ -0,0 +1,6 @@
+resource "aws_eks_addon" "coredns" {
+  cluster_name      = "lead"
```
ChrisSchreiber, line 2: Should this be a variable? Not all our clusters are called "lead"

ChrisSchreiber (changes requested):

modules/environment/aws/eks/addons.tf
mrparkers: yes, we initially pulled this from the eks module's cluster ID, but when we tested this in sandbox (while it was still broken), it wouldn't allow us to proceed, so we temporarily hardcoded it.  I reverted that change so we should be good to go here.

silver-surfer94 (approved):

PR #582, opened 2022-01-28
fix: add permission boundary to sparky sa role

ChrisSchreiber:


.github/workflows/terraform-validate.yml
```diff
@@ -79,6 +79,7 @@ jobs:
         if: ${{ always() }}
         with:
           github_token: ${{ secrets.github_token }}
+          tfsec_version: "v0.63.1"
```
mrparkers, line 82: is this the current version the pipeline was using?  or are you pinning it to an older version?

ChrisSchreiber: 0.63.1 is what it used the last time is passed... today it used 1.0.0 which changes a bunch or rules
try `docker run --rm -it -v "$(pwd):/src" aquasec/tfsec /src -e general-secrets-sensitive-in-variable,general-secrets-sensitive-in-local,general-secrets-sensitive-in-attribute` to see changes

mrparkers: thanks for the context!

mrparkers (approved):

PR #584, opened 2022-02-01
feat: deploy new version of sparky

mrparkers:
this also does some refactoring by removing the `sdm` module - I didn't think it was super necessary since it's only creating a few resources and it isn't used elsewhere.  this helped reduce the confusion with how we map variables to the underlying operator-toolchain chart (even though we'll be removing that eventually anyways).


ChrisSchreiber (commented):
...

mrparkers:
Most of our modules now are nice and small, they do one thing and one thing well, and it's really clear what variables they expect and how they're used.  So I wouldn't want to change any of those.

The `sdm` module, on the other hand, is really cluttered.  I understand that it being concerned with multiple things is really more of a reflection of the `operator-toolchain` pattern we used to use, but it is still clunky to use, and the variables aren't well defined at all.  For example, a third of the variables we send to this module are actually nested within a `product_vars` map, which we just pull them out of when creating the chart.  It doesn't really make sense, and as a result, this module looks nothing like our other purpose-built modules that you'd find within the `modules/tools` folder, for example.

I didn't feel like it was worth the time to refactor this module since we'd eventually no longer need the `operator-toolchain` chart at all.  In the future, I think we could package `sparky` and `product-operator` into their own modules that we could use from within this file.  I could start that process now, but I needed some stuff that was originally within the `sdm` module (the image pull secret, for example), and I really didn't want to contribute to the bloat we already have within that module.

Maybe we can chat in person and work on a compromise for this though.

stages/apps/lead/variables.tf
```diff
@@ -56,6 +56,14 @@ variable "opa_failure_policy" {
   default = "Fail"
 }
 
+variable "enable_sparky" {
+  default = true
```
ChrisSchreiber, line 60: I want us to start using a disabled by default approach to feature flags. It makes it easier to see what is running in an environment by just looking at the inputs in the terragrunt.hcl instead of having to check what is enabled by default.
```suggestion
  default = false
```

mrparkers: that makes sense to me

ChrisSchreiber (approved):

PR #586, opened 2022-05-06
BREAKING CHANGE: updated eks module to 18.20.5 from 17.18.0

anthonygauthier:


mrparkers:
Overall this looks good, but I'm confused as to why we're calling this a breaking change, and we'll be cutting a v3 release if this is merged.  I don't think we need to cut a breaking change every time we bump one of our dependencies to a new major version.

anthonygauthier:
@mrparkers Fair enough, I think I might actually have confused myself in the sense that their upgrade from v0.17.X to v0.18.X. was breaking. But it's true that for us, there's no breaking change if the migration is done properly. Good catch! I'll amend the commit.

mrparkers:
I think their module update was probably breaking given some of the changes I saw you made.

Anyways, this looks good at a glance, although I'm in an airport so I don't have the capacity to review in further detail.

mrparkers (approved):

djtjwillia:
Looks fine to me, we already know there could be issues in the deployed apps in the cluster with the pending upgrade, so I think this is fine. 

We do need to get all the validations passing.

PR #588, opened 2022-05-10
fix: using count as a conditionnal in code-services for bucket

anthonygauthier:


stages/cloud-provider/aws/lead/codeservices.tf
```diff
@@ -1,4 +1,5 @@
 module "codeservices" {
+  count  = var.enable_aws_code_services ? 1 : 0
   source = "../../../../modules/environment/aws/code-services"
 
   enable_aws_code_services    = var.enable_aws_code_services
```
ChrisSchreiber, line 5: do we still need to pass the flag into the module?

ChrisSchreiber (changes requested):
Looks good. One small change

PR #595, opened 2022-07-15
fix: add exclusion on elasticsearch-curtor-values.tpl

chodges7:
We want to include a filter on the delete old indices action that excludes the github actions metrics indices. We want to exclude these repos because we want historical data.


chodges7:
We swapped the filters around to future-proof the `delete-old-indices` action. We want to delete the logging indices `kubernetes_cluser` and `jaeger-jaeger` since they contain a lot of usage data that isn't needed after 14 days.

ChrisSchreiber (approved):

PR #601, opened 2022-07-25
fix: add resource arn to iam policies

chodges7:
This PR resolves a batch of TFSec errors that disallowed wildcards on which resources can be affected by a policy.

### DIFF
--- a/modules/environment/aws/iam/cert-manager/main.tf
+++ b/modules/environment/aws/iam/cert-manager/main.tf
@@ -26,7 +26,6 @@ resource "aws_iam_role" "cert_manager_service_account" {
 EOF
 }
 
-#tfsec:ignore:aws-iam-no-policy-wildcards
 resource "aws_iam_role_policy" "cert_manager" {
   name = "${var.cluster}-cert-manager"
   role = aws_iam_role.cert_manager_service_account.name
--- a/modules/environment/aws/iam/cert-manager/main.tf
+++ b/modules/environment/aws/iam/cert-manager/main.tf
@@ -34,10 +34,17 @@ resource "aws_iam_role_policy" "cert_manager" {
     {
       "Version" : "2012-10-17",
       "Statement" : [
+        # Cert-Manager can be limited in the access it has for each hosted zone that it touches,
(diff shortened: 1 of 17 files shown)


modules/environment/aws/iam/cert-manager/main.tf
```diff
@@ -26,34 +26,27 @@ resource "aws_iam_role" "cert_manager_service_account" {
...
-    }
- ]
-}
-EOF
+      "Version" : "2012-10-17",
+      "Statement" : [
+        {
+          "Effect" : "Allow",
+          "Action" : "route53:GetChange",
+          "Resource" : "${formatlist("arn:aws:route53:::change/%s", var.hosted_zone_ids)}"
```
mrparkers, line 40: this doesn't look right.  I would expect this to be a "change ID", not a hosted zone ID.  I think this needs to be left as a wildcard

mrparkers (changes requested):

modules/environment/aws/iam/cert-manager/main.tf
```diff
@@ -26,34 +26,34 @@ resource "aws_iam_role" "cert_manager_service_account" {
...
-        "Action": "route53:ListHostedZonesByName",
-        "Resource": "*"
-    }
- ]
-}
-EOF
+      "Version" : "2012-10-17",
+      "Statement" : [
+        # Cert-Manager can be limited in the access it has for each hosted zone that it touches,
+        #   but according to the docs[1], "cert-manager needs to be able to add records to
```
mrparkers, line 38: I think the wildcard is needed specifically to get a change to a record set that was already requested.  The permission below this one is for `ChangeResourceRecordSets`, which would be more in line with what your comment is describing.

In other words, this comment isn't wrong, but I don't think it explains why the wildcard is necessary.  There's no way for us to know the change ID when we apply this terraform, because those changes are happening at runtime.

chodges7: Ahh okay. I didn't see a very explicit explanation on the docs page that I found so this makes a lot of sense.

mrparkers (approved):

PR #606, opened 2022-08-01
feat: add encryption and public access block to S3 buckets

Micahperez2:


modules/environment/aws/code-services/main.tf
```diff
@@ -22,21 +22,50 @@ data "aws_subnets" "eks_workers" {
...
   bucket = aws_s3_bucket.code_services_bucket.id
   versioning_configuration {
     status = "Enabled"
   }
 }
 
+resource "aws_s3_bucket_logging" "code_services_logging" {
+  bucket = aws_s3_bucket.code_services_bucket.id
+
+  target_bucket = "code-services-${var.account_id}-${var.cluster}"
```
ChrisSchreiber, line 56: Sending logs to the source bucket is not recommended https://aws.amazon.com/premiumsupport/knowledge-center/s3-server-access-logs-same-bucket/

We should probably setup a dedicated s3 bucket just for logging.

ChrisSchreiber (changes requested):

modules/environment/aws/code-services/variables.tf
```diff
@@ -5,3 +5,4 @@ variable "cluster" {}
 variable "openid_connect_provider_arn" {}
 variable "openid_connect_provider_url" {}
 variable "vpc_name" {}
+variable "s3-logging-id" {}
```
ChrisSchreiber, line 8: We should keep the variable naming consistent here and use underscores throughout 
```suggestion
variable "s3_logging_id" {}
```

ChrisSchreiber (changes requested):

ChrisSchreiber (approved):

PR #610, opened 2022-08-08
fix: resource limit and request added to autoscaler to resolve crash loop issue

azizn03:


ChrisSchreiber:
Part of the pipeline checks that they PR title uses conventional commit syntax https://www.conventionalcommits.org/en/v1.0.0/ You will need to rename the PR for the checks to pass. Let me know if you want some background info on conventional commits.

ChrisSchreiber (approved):

PR #615, opened 2022-08-10
feat: velero daily backup schedules

Spots87:


Spots87:
About this test. Since harbor has the resource for the velero crd and there isnt a way (at least that I see) to implement a dependency in the harbor/basic_test "enable_velero" was set to false for purpose of the test. Otherwise it will fail because the crd does not exist. In actual terraform testing velero_status = try(module.velero[0].velero_status, true) in combination with depends_on = [ var.velero_status ] causes the resource to wait for Velero to be deployed which creates the schedules.velero.io crd.

modules/tools/kube-prometheus-stack/main.tf
```diff
@@ -42,4 +42,79 @@ resource "helm_release" "prometheus_resources" {
...
+        "ttl"             = "72h0m0s"
+        "volumeSnapshotLocations" = [
+          "aws-s3",
+        ]
+      }
+    }
+  }
+  depends_on = [
+    helm_release.kube_prometheus_stack
+  ]
```
ChrisSchreiber, line 81: is there a reason all the other schedule resources depend on velero_status except this one?

modules/tools/elasticsearch/main.tf
```diff
@@ -77,3 +77,42 @@ resource "helm_release" "elasticsearch_curator" {
...
+          "aws-s3",
+        ]
+      }
+    }
+  }
+  depends_on = [
+    helm_release.elasticsearch,
+    var.velero_status
+  ]
+}
```
ChrisSchreiber, line 118: You should try setting up your editor to and new lines to the end of files to avoid this red mark :) https://stackoverflow.com/questions/44704968/visual-studio-code-insert-newline-at-the-end-of-files

modules/tools/elasticsearch/main.tf
```diff
@@ -77,3 +77,42 @@ resource "helm_release" "elasticsearch_curator" {
     helm_release.elasticsearch
   ]
 }
+
+resource "kubernetes_manifest" "schedule_velero_elasticsearch_daily_backup" {
```
ChrisSchreiber, line 81: Maybe we should make the schedule resource its own modules to set some defaults and reduce duplicate code?

ChrisSchreiber (changes requested):

Spots87:
Reworking to use velero-schedule module and resolve CRD dependency issue.

PR #627, opened 2022-09-23
feat: add iam role for aws-terraform runner

ChrisSchreiber:


steve-liatrio (approved):
I don't have much context but I don't see any issues

